Issue #6, opened 2018-11-27
Panic on package build failure.


c-edw:
I am trying to install some packages and am getting a panic on `src/wrapped.rs:61:2`. The backtrace is pretty unhelpful, but it looks like the command that was being executed didn't return 0. 

Ideally this shouldn't panic, but should just print a warning or error.

vn971:
@c-edw Do you have backtrace at all, or is it just unhelpful? If you're compiling via AUR, you need to uncomment `options+=(!strip)`, because otherwise `makepkg` strips out all debug symbols.

vn971:
@c-edw yes, panicking is indeed not a nice thing to have. I need to make an iteration of cleaning them up again, probably.

By the way, have you installed the dependencies?: https://github.com/vn971/rua#install-dependencies

c-edw:
The first time, I followed `Install (the AUR way)`, which lead to this issue. I've just tried `Install (the Rust way)` but get an immediate panic when trying to install something. I'm the `root` user in a Docker container, if that matters.

```
thread 'main' panicked at 'Failed to git-clone repository wlroots-git: Os { code: 2, kind: NotFound, message: "No such file or directory" }', libcore/result.rs:1009:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::panicking::default_hook::{{closure}}
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
             at libstd/panicking.rs:211
   2: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:227
             at libstd/panicking.rs:477
   3: std::panicking::continue_panic_fmt
             at libstd/panicking.rs:391
   4: rust_begin_unwind
             at libstd/panicking.rs:326
   5: core::panicking::panic_fmt
             at libcore/panicking.rs:77
   6: core::result::unwrap_failed
   7: rua::aur::fresh_download
   8: rua::wrapped::prefetch_aur
   9: rua::wrapped::install
  10: rua::main
  11: std::rt::lang_start::{{closure}}
  12: main
  13: __libc_start_main
  14: _start
```

c-edw:
I have installed the dependencies also.

vn971:
I see that it's in `rua::aur:fresh_download` then.
I'll check this method, as well as other usages of `unwrap` that I have.
[...]

PR #7, opened 2018-11-27
Add git to dependencies.

c-edw:
Resolves one sub-issue in #6.


Morganamilo:
Why even encourage subverting the package manager anyway? Why not remove the section entirely and have the pkgbuild be the only supported install method?

vn971:
@Morganamilo good question. Though on the other hand, this may help people coming from Rust understand what's actually going on.

vn971:
@c-edw thanks for the PR! I'll add "git" to dependencies as well, as RUA uses it to fetch packages anyway.

PR #10, opened 2019-01-21
Clippy fixes

senden9:
Hi!

I want to implement a search with results in the terminal. For that I first fixed all clippy lints as seen in the first commit of this PR. The most "fixes" are just [expect_fun_call](https://rust-lang.github.io/rust-clippy/master/index.html#expect_fun_call) things. So you basically save one `format` call that is unnecessary the most time.

Let me know how you think about that changes!

### DIFF
--- a/src/aur_download.rs
+++ b/src/aur_download.rs
@@ -24,18 +24,18 @@ fn assert_command_success(command: &Output) {
 pub fn fresh_download(name: &str, dirs: &ProjectDirs) {
 	lazy_static! {
 		static ref name_regexp: Regex = Regex::new(r"[a-zA-Z][a-zA-Z._-]*")
-		.expect(&format!("{}:{} Failed to parse regexp", file!(), line!()));
+		.unwrap_or_else(|_| panic!("{}:{} Failed to parse regexp", file!(), line!()));
 	}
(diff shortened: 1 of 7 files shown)


vn971:
// Preliminary comment. Thanks! I'll return to it tomorrow or on the week.

vn971:
// Added the check as part of CI as well. For both rustfmt and clippy.

senden9:
I did them by hand. I think i heard from some `rustfix`-tool or something like that that can do things like this automatically. But when it comes to codebase changes like this i review the things "by hand" and look if the changes make sense. From time to time clippy has false positives.

vn971:
I see. Thanks a lot for the contribution!

vn971:
BTW, I noticed that some errors are swallowed in the newer rewrite if we use:
`result.unwrap_or_else(|_| panic!("")`

Something that `expect` handled better actually, because it has shown the underlying reason. I think this re-write is OK though:
`result.unwrap_or_else(|err| panic!("blah blah blah, err: {}", err)`

Didn't play out with yet, especially regarding how exactly the error would be displayed, but I think it should work. I've raised a PR with the change: https://github.com/vn971/rua/pull/23

PR #17, opened 2019-03-09
Restructure imports for rust 2018

Morganamilo:


vn971:
Hi, thanks for the PR! Looking at it now. The travis job probably needs improvement, but what it means, is that you have to write yourself as contributor inside Cargo.toml :)

I'll take a look at other parts meanwhile.

vn971:
Like this, with the same credentials as in git itself, if that's no problem for you.. https://github.com/vn971/rua/blob/master/Cargo.toml#L7

Morganamilo:
Hmm kinda a weird requirement. Personally I don't think every minor contributor should be listed there. But your project :P

vn971:
I wonder though why the formatting changed. Previously, IIRC rustfmt had split imports like `something::{a, b, c}` into separate lines (which makes sense as it simplifies `git diff`). Wonder why if something changed in that regard.

Morganamilo:
I never heard of rustfmt changing the way imports are structures I guess it's just personal preference. Up to you if you prefer it or not.

vn971:
@Morganamilo interesting. Well, regardless, if tests pass, it's OK by me for now.
I don't have any real personal preferences on this, yet.

Personal current TODO is to somehow calculate upstream package diff, so that the user would have more confidence in packages being installed. Otherwise I have to reall re-read whole repo each time, and it's absolutely not convenient, and forces you to skip doing it..

vn971:
Thanks! Merged. I'll look into this CI thing separately, somewhy it didn't work as it should.

Morganamilo:
My local git config and the name I have set on github are different. That may be while it's failing.

vn971:
Theoretically, it should read `git log` to get the username. In practice, travis-ci probably disabled this for some security reasons. Anyway, I just disabled this check to avoid confusion -- it had low value anyway.

PR #22, opened 2019-03-10
Use the srcinfo crate for srcinfo parsing

Morganamilo:
This removes quite a couple of lines from rua while fixing a bug and altering the behaviour a little bit (see the commit messages)

This is also my own crate, I feel a little weird about pushing my own stuff on others but it does seem to meet your needs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -334,6 +334,7 @@ dependencies = [
  "log 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 1.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "rm_rf 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "srcinfo 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "tar 0.4.21 (registry+https://github.com/rust-lang/crates.io-index)",
  "uname 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "users 0.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -343,6 +344,11 @@ dependencies = [
 [[package]]
 name = "smallvec"
 version = "0.6.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "srcinfo"
+version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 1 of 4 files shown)


vn971:
Thanks for the PR! This will may take some time to read through carefully. One thing to note early is that the compilation is currently failing.

Also, there are PKGBUILD settings like "source_x86_64", not only literally "source" (as written in a git comment). Also hashes matter and would better be checked straight away, to avoid potentially dangerous incorrect files being left in the filesystem. It could be though that everything is correct already -- I haven't read it all through yet.

vn971:
UPD: and I've fixed `shellcheck` in CI (.travis.yml), should be OK now.

Morganamilo:
>Also, there are PKGBUILD settings like "source_x86_64", not only literally "source" (as written in a git comment).

Thats already handled here https://github.com/vn971/rua/blob/d6605510ec63969c164f3869db203655d02b80ae/src/srcinfo.rs#L30-L36

> Also hashes matter and would better be checked straight away.

The current code is already using `--skipinteg` but if you want me to change that and check them right away I'll do that.

vn971:
Ah, see that now (`source`), thanks for explaining.

Yeah, I think it could make sense to check integrity straight away, to be on the safe side.

Morganamilo:
Done and also tweaked the iter chain slightly. Passes the checks for once too :P

vn971:
Hey! Sorry for the long review -- I wanted to read through the crate carefully, and it took quite some time. The crate is pretty through!
I also wanted to adapt a few things, check out the latest commit if interested: https://github.com/vn971/rua/commit/e59bf1026f598ec07ec512e21c3bea5da4af5040

BTW, I would it be a good idea to ask your review on new features I'm implementing? Would that be interesting for you? To track changes, and to have proper time to raise questions/concerns if you'll have any.

Morganamilo:
I enjoy contributing to projects, Especially AUR stuff and am looking at getting into rust. So in that aspect yes. But I'm not really interested in jailing or any of that sort of stuff.

So if you want my opinion on features I'm probably gonna just push towards chroot builds and local repos. Although I'm happy to comment on the technical side of things. I'm pretty experienced with the AUR at this point.

vn971:
I see. Interesting to know!
Regarding interest in Rust but no so much jailing -- I guess that's fine, if the two do not contradict each other and usability (probably they don't).

Regarding local repos -- I haven't really considered it yet. Might be something to look in in the future. It was not on top of my personal priorities yet as I wanted to finish AUR JSON API first.

I'll assign some PR's to you now and then then -- feel free to discard those you don't like though.

Morganamilo:
For the RPC there's https://github.com/rusty-crates/aur.rs and https://gitlab.com/davidbittner/raur. I assume you know about the former since you opened an issue there.

The latter once seems much simpler to use and the code is more simple.

But then the former extents reqwest/hyper via traits which I guess may be the more rust like approach? This latter is probably better too if you want to split up requests asynchronously (which you probably do due to the 4kb url limit).

Although assuming all my PRs eventually get merged they should both have an almost identical api anyway.

PR #23, opened 2019-03-17
improve error logging for Result.unwrap_or_else

vn971:


vn971:
@Morganamilo pretty boring PR I guess.🤷‍  I hope to push some more interesting PR-s in the future!

vn971:
BTW, if any particular PR seems boring enough and you don't want to read it through -- you don't have to, just tell and I'll skip it (merge manually).

vn971:
Regarding `line!()` and panicking -- rust only gives really useful stack traces if debugging symbols aren't stripped off. They are, however, stripped by `makepkg` by default. If the user builds "rua" package as-is, stack trace will be very limited and line numbers won't be visible. There are some language proposals in Rust that would ease up this `file!()` and `line!()` nonsense, but currently that's what you must do if debugging symbols are stripped off and you want line numbers, apparently..

I'll think more about the nicer error formatting -- good question indeed.

Morganamilo:
That makes sense then. Also failure has a back trace feature. Not sure if it's effected by stripping though.

vn971:
Thought of addressing the linked issue ("better logging") along the way, but let's do it consecutively.

vn971:
merged.

PR #24, opened 2019-03-18
Use &Path over &PathBuf

Morganamilo:
like str and String, Path and Pathbuf are borrowed/owned counterparts.
&Path should always be used over &Pathbuf.

This is based on #23 so the diff is a mess right now. Merge that first before reviewing this.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,7 +12,7 @@ mod wrapped;
 use std::fs::{File, OpenOptions, Permissions};
 use std::io::Write;
 use std::os::unix::fs::PermissionsExt;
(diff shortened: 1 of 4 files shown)


vn971:
Thanks! Merged.

vn971:
Maybe I should use the GH button to have the pull request status as "merged".. IDK why it's not auto-detected.

vn971:
I think I didn't really merge this branch in Mar 19 though... Making the same change on current codebase and formally "merging" this PR branch.
Feel free to delete the "path" branch now @Morganamilo , I think it has been truly merged now.

PR #26, opened 2019-03-22
Use checkdepends when resolving deps

Morganamilo:


vn971:
Indeed, it's a poper change.
Thanks!

Issue #28, opened 2019-03-23
Frustrated: confusing behavior when installing xen


b1f6c1c4:
I enabled `multilib` from pacman first and ran `rua install xen`
After a very looooooooong wait the compilation is finished.
`rua` asked me
```
Package /home/b1f6c1c4/.cache/rua/xen/build/target/xen-4.11.0-2-x86_64.pkg.tar.xz has no SUID files.
[I]=show install file, [E]=list executable files, [L]=list all files, [T]=run shell to inspect, [O]=ok, proceed.
```
And I typed `i` to view the install file. It looks nice. Afterwards,
```
Package /home/b1f6c1c4/.cache/rua/xen/build/target/xen-4.11.0-2-x86_64.pkg.tar.xz has no SUID files.
[I]=show install file, [E]=list executable files, [L]=list all files, [T]=run shell to inspect, [O]=ok, proceed. o


Package /home/b1f6c1c4/.cache/rua/xen/build/target/xen-docs-4.11.0-2-any.pkg.tar.xz has no SUID files.
[E]=list executable files, [L]=list all files, [T]=run shell to inspect, [O]=ok, proceed. o

Packages need to be installed:

    pacman -U --needed /home/b1f6c1c4/.cache/rua/xen/checked_tars/xen-docs-4.11.0-2-any.pkg.tar.xz

Enter S to `sudo` install it, or install manually and press M when done. s
loading packages...
resolving dependencies...
looking for conflicting packages...

Packages (1) xen-docs-4.11.0-2

Total Installed Size:  3.94 MiB

:: Proceed with installation? [Y/n]
(1/1) checking keys in keyring                     [######################] 100%
(1/1) checking package integrity                   [######################] 100%
(1/1) loading package files                        [######################] 100%
(1/1) checking for file conflicts                  [######################] 100%
(1/1) checking available disk space                [######################] 100%
:: Processing package changes...
(1/1) installing xen-docs                          [######################] 100%
:: Running post-transaction hooks...
(1/1) Arming ConditionNeedsUpdate...
Packages need to be installed:

    pacman -U --needed /home/b1f6c1c4/.cache/rua/xen/checked_tars/xen-docs-4.11.0-2-any.pkg.tar.xz

Enter S to `sudo` install it, or install manually and press M when done. s
loading packages...
warning: xen-docs-4.11.0-2 is up to date -- skipping
 there is nothing to do
Packages need to be installed:

[...]

PR #29, opened 2019-03-25
remove cache directories later

vn971:


vn971:
This will fix early cache removal issue from https://github.com/vn971/rua/issues/28

vn971:
Sorry for pushing more commits after the PR was already opened. Understood that I want to fix a few things in same PR first.

A better and/or long-term solution would probably not to mess up with PKGDEST and the "target" subdirectory at all. For that RUA needs to teach itself how to deal with split packages and versions first though. So currently the intent is to wipe out whole dirs if it can help avoiding conflicts/mismatches.

vn971:
@Morganamilo hi, did you have any thoughts on the PR-s yet? If there's a chance that you'll have some time on the week, I'd wait. If you know you won't have any, I'd merge.
Thoughts?

Morganamilo:
Assuming the title is correct why remove them later? Why not just not remove them at all?

vn971:
@Morganamilo because there is no built-in automated way to upgrade packages, at least not yet. Having clean build directory is what I'd like to have even for future. E.g. even if we show upstream git diff, I would still prefer to copy the edited AUR repo to a separate dir and build there cleanly. Does that make sense to you?

Morganamilo:
That's fine enough then I guess. I don't really know how rua builds things in the first place so there's not much point in me reviewing. And like I said I hate the official approval stuff.

Issue #8, opened 2018-11-29
Support insecure builds | docker


vn971:
Currently, RUA does not support building packages without the namespace jail (bubblewrap). Unfortunately, it does not work in e.g. docker.

Adding support for insecure (jail-less) builds could help users of docker, see e.g. https://github.com/vn971/rua/issues/6#issuecomment-442188061

vn971:
If anybody thinks it's very important and wants to write a PR for that, I think it could be merged presuming:
* it's a quality PR in terms of code and maintenance
* the feature itself is hidden under a verbose-only warning flag

Personally, I see too few evidence on needing that yet, so not willing to work on it yet.

gwpl:
```
bwrap: capset failed: Operation not permitted
2022-10-13 09:37:29 [ERROR] - Failed to run bwrap.
2022-10-13 09:37:29 [ERROR] - A possible cause is if RUA itself is run in jail (docker, bwrap, firejail,..).
2022-10-13 09:37:29 [ERROR] - If so, see https://github.com/vn971/rua/issues/8
```

gwpl:
cross-linking: I've started thread also on Arch Linux forum: https://bbs.archlinux.org/viewtopic.php?pid=2063741#p2063741

PR #30, opened 2019-03-25
check bwrap availability

vn971:
In case of error, exit with a nicer message

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -97,11 +97,11 @@ fn main() {
 	if !Command::new("bwrap")
 		.args(&["--ro-bind", "/", "/", "true"])
 		.status()
-		.expect("bwrap binary not found. RUA uses bwrap for security isolation.")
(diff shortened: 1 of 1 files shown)


Morganamilo:
Not sure what the point of this is. Seems like this is what `depends=()` is for.

vn971:
@Morganamilo no, it's meant as a different thing. `depends=()` checks package availability. Some files lying around on filesystem.

This code, however, is meant to test whether it actually works. Like in docker https://github.com/vn971/rua/issues/8

Easiest way to test:
`bwrap --dev-bind / / rua install test`
The command `rua install test` will then be run inside a jail. It wil fail at run-time. With this PR, there'll be a clear explanation why it fails.

Morganamilo:
Oh right I skimmed it and only saw the `bwrap binary not found` message.

vn971:
@Morganamilo yes, that's exactly what I'm expecting. (Wouldn't mind anything else, but I'm realistic.:)

PR #32, opened 2019-05-05
Use my branch of raur

Morganamilo:
With this PR `rua search test` works.

This is also the first time I've tested search and i'm not a fan of the output format. But to keep this PR focused all it does is switch the branch and nothing more.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -317,16 +317,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "display_derive"
-version = "0.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "quote 0.3.15 (registry+https://github.com/rust-lang/crates.io-index)",
- "syn 0.11.11 (registry+https://github.com/rust-lang/crates.io-index)",
- "synstructure 0.6.1 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
-[[package]]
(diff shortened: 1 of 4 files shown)


vn971:
@Morganamilo I cannot publish it on crates.io though. I get this:
```
error: crates cannot be published with dependencies sourced from a repository
either publish `raur` as its own crate and specify a version as a dependency or pull it into this repository and specify it with a path and version
(crate `raur` has repository path `https://gitlab.com/morganamilo/raur?rev=11ac81acaf03995151e872d6fa349eae2e86b3dc`)
```

Morganamilo:
>Only change I do not understand so far is to why change method search to search_by in raur library

`search` is a convenience function and uses the aur's default of search by name-desc. While `search_by` lets you specify what to search by.

Typically you'd use overloading for this but rust doesn't support it.

Also may as well add. I don't really get why you have it set to search only by name and not name desc. But as I mentioned above this PR is purely to change the dep. And not functionality.

vn971:
@Morganamilo but both AUR-s `search` RPC command and raur-s current `search` require you to specify the `search_by` thing explicitly. They all work as "search_by", just the AUR version is named without the "_by".

Morganamilo:
In my branch `search` defaults to `name-desc` though. So don't really understand what you're trying to say here.

vn971:
Oh. Got it.

PR #35, opened 2019-05-21
cargo upgrade

Morganamilo:
rua:
    Upgrading raur v0.3.0 -> v1.0.1
    Upgrading colored v1.8 -> v1.8.0
    Upgrading tar v0.4.25 -> v0.4.26


Morganamilo:
Thanks to the backwards compatibility we don't need to change out raur code. So just `cargo upgrade` is all.

PR #36, opened 2019-05-22
[DRAFT] Much improve show command

Morganamilo:
Here is my take at the `show` command (which I would recommend renaming to `info` btw)

Months ago I wrote a small command line tool that prints AUR info. I pretty much just copy pasted the code into rua.

So as it is, this code doesn't fit in too well, adds a bunch of new dependencies. But if you like this approach then it can spread the the rest of rua. I'd ideally also replace the search code with a similar to this approach instead of using prettytable.

Also eventually if this code becomes mature I may move it into it's own crate. But for now I'll inject it into rua directly so changes can happen fast.

This code also properly wraps and handles colour:
image

image

image


Morganamilo:
Whoops this was meant to be a draft. This is a proof of concept more than something I want merged right now.

vn971:
@Morganamilo what do you plan to do about "search" command though? This is the most complex one I think, and it's the only one using `prettytable` now.

With regards to "info", I just threw in something very simple for now, just to show the basic data. Improvements should conceptually be straightforward, but I of course appreciate adding any data that you want to have in the view.

Morganamilo:
>@Morganamilo what do you plan to do about "search" command though? This is the most complex one I think, and it's the only one using prettytable now.

Just like how I've done with info. Copy pacman's layout. Most AUR helpers already do this.

Morganamilo:
I'm looking for feedback so what I really want to know is:

- Are you happy with the formatting?
- Are you happy with the approach for colouring as well as the way they are enabled/disabled?
- Are you happy with the dependencies?

vn971:
> happy with the formatting?

I didn't see the `search` formatting, I can't tell without that! All the screenshots above are for `show`, not `search`. Since you're saying about "pacman layout", I've made some screenshots of it:
pacman:
pacman
rua:
rua

Do you really like pacman-s output more?! I can barely read the packet-s names with it! It's always a pain for me to distinguish that from the descriptions.

> happy with the approach for colouring
You mean the bold text, right? I'm fine with the way you used bold text for marking "keys", and normal text for data. Date format and the way to show dependencies might be OK as well, but we will probably want to distinguish AUR and non-AUR deps in the future. I'll have to read a bit more code to answer on the code aspect of disabling/enabling, give me a some hours or weekend on that.

> the dependencies
it'd be good of course to pick either `ansi_term` or `colored`, and use it everywhere. Not strictly necessary, but still nice. By the way, colored answers the question of having or not having colors pretty straightforwardly, no? Do you know any downsides of how it uses it?

You can also always reach me (and other people) on IRC, feel free to use whichever's more comfortable.

Morganamilo:
> I didn't see the search formatting,

There is no search yet. As the title says the code is just for info.

>Do you really like pacman-s output more?!

Definitely, especially since I always have colour enabled and that helps break up the parts.

See yay's output which also includes votes, popularity, install status and groups

image

Hard to fit all that into a table.

>You mean the bold text, right?

By approach I meant meant on the code side, the use of ansi_term + atty and the way the config struct stores the colour settings.

>it'd be good of course to pick either ansi_term or colored, and use it everywhere.

This is part of what I was talking about with:

>Months ago I wrote a small command line tool that prints AUR info. I pretty much just copy pasted the code into rua.

>So as it is, this code doesn't fit in too well, adds a bunch of new dependencies. But if you like this approach then it can spread the the rest of rua.

Morganamilo:
Forgot to reply to this bit whoops:

>By the way, colored answers the question of having or not having colors pretty straightforwardly, no? Do you know any downsides of how it uses it?

It does? I remember trying it a bit before but couldn't find out a good way to do it.

vn971:
That's strange: I find RUA-s `search` output way more readable. Especially if made on a 1080p monitor with a font that I normally use. I probably need to include ratings for it to be completely fair (+8 characters), but as of current output:
yay: 
rua: 


I guess I'll need input from other people as well to decide on `search` specifically. With `info`, I'm fine with the proposal on the looks.

> It does? I remember trying it a bit before but couldn't find out a good way to do it.

It respects the `NO_COLOR` and `CLICOLOR` environment variables: https://bixense.com/clicolors/

vn971:
UPD: apparently `colored` doesn't follow the specs in respecting the `tty` thing. Maybe I'll submit a PR to them..

vn971:
Is this still a draft, or is it intended to be merged BTW? I like, for example, the usage of kebab-case.

Regarding the other branch, it still needs to be looked at properly..

Morganamilo:
It will be ready to merge when I also get search to use the same system and get the entire project using the same dependencies. Instead of how it's using two different colour libs right now.

You didn't seem to take too much of a liking to this so I'm waiting on you get the colours sorted out how you want them first.

src/config.rs
```diff
@@ -0,0 +1,86 @@
+use ansi_term::Color::Red;
+use ansi_term::Style;
+use atty::Stream::Stdout;
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub struct Config {
```
vn971: I'm not sure the name is ideal for our use case as we might start having "rua config" in the future. Also, it truly isn't a config -- it's CLI arguments.

src/config.rs
```diff
@@ -0,0 +1,86 @@
...
+		long = "color"
+	)]
+	pub color: Colors,
+	#[structopt(subcommand)]
+	pub action: Action,
+}
+
+#[derive(StructOpt)]
+#[structopt(rename_all = "kebab-case")]
+pub enum Action {
```
vn971: following `clap` terminology, we can call it `Command` or `Subcommand`.

src/config.rs
```diff
@@ -0,0 +1,86 @@
...
+		#[structopt(help = "Archive to check", required = true)]
+		target: PathBuf,
+	},
+}
+
+fn parse_color(s: &str) -> Colors {
+	match s {
+		"auto" if atty::is(Stdout) => Colors::new(),
+		"always" => Colors::new(),
+		_ => Colors::default(),
```
vn971: I think it'd be a bit nicer from type system perspective to consider `"never" => Colors::default(),` and `panic` for the rest. Or better yet, create enum to hold the value and get parsing for free. There should be a way to automatically parse enum in `structopt` right?

src/fmt.rs
```diff
@@ -0,0 +1,52 @@
...
+                    pos += word.len() + 1;
+                }
+            }
+        }
+        _ if list => print!("{}", v.collect::<Vec<_>>().join("  ")),
+        _ => print!("{}", v.collect::<Vec<_>>().join(" ")),
+    }
+
+    println!();
+}
```
vn971: Very cool idea&design to handle word splitting! :heart:

I wonder if it'll even be valuable outside of RUA once we polish it here. The current code doesn't support the more general use case of many columns BTW, right? E.g. the `search` command. But here I'm just curious, it's not blocking the merge.

src/info.rs
```diff
@@ -0,0 +1,87 @@
...
+    let mut hm = HashMap::with_capacity(pkgs.len());
+
+    for chunk in pkgs.chunks(150) {
+        let pkgs = raur::info(&chunk)?;
+        for pkg in pkgs {
+            hm.insert(pkg.name.clone(), pkg);
+        }
+    }
+
+    let mut info = Vec::with_capacity(hm.len());
```
vn971: hm?)
OK, jokes aside, maybe rename it to a longer name, like `details` or `name_to_pkg` or something?

src/config.rs
```diff
@@ -0,0 +1,86 @@
...
+
+#[derive(Default)]
+pub struct Colors {
+	pub field: Style,
+	pub error: Style,
+}
+
+impl Colors {
+	pub fn new() -> Colors {
+		Colors {
```
vn971: As you proposed in IRC some time ago, I would indeed let this property set the environment to enforce colors (or lack thereof). And openly document the behavior. At least, do you see any downsides to doing so? You seem to be kinda opinionated and very knowledgeable here @Morganamilo.

src/config.rs
vn971: Though I'm not sure why I'm writing this, "action" is probably nice as well. Nevermind.

vn971:
> the show command (which I would recommend renaming to info btw)

I thought about it as well, but what stopped me so far:
* in Debian world, the command is named `show`, e.g. `apt show something`. In a build tool I (SBT) of a language I use (Scala), there's a `show` command as well, but no `info`.
* if we use `info`, then we'll have a 2-letter common prefix with another command, `install`. E.g. now you have to type `rua s<TAB> name` and `rua i<TAB> name`, but with the change you'll have to type `rua inf<TAB> name` and `rua ins<TAB> name`.

vn971:
Thanks for the cool column wrapping code and pushing the colors question -- merged!

Morganamilo:
The thing is info is already an established convention. Thanks to `pacman -Si` and the RPC request having `type=info` pretty much all helpers have adopted the word info for this.

Auracle also has it's on `show` command which prints the pkgbuild to stdout. I was thinking of eventually implementing this here too.

vn971:
Ah indeed, `pacman -Si` can even be written as `pacman -S --info` :thinking:
OK, I guess we can change that one.)

Pushed to master.

Issue #9, opened 2019-01-13
depends_arch should not assume pacman architecture == uname


Morganamilo:
For architecture specific fields (e.g. depends_x86_64) the architecture should correspond to the architecture defined in pacman.conf, not the system architecture.

On x84_64 this doesn't really matter as pacman.conf and uname happen to agree.

But on other architectures like arm, `uname -m` gives `armv7l` while pacman.conf contains `Architecture = armv7h`.

vn971:
Remark to anyone fixing this bug (probably me).
The correct way to get architecture is to invoke `pacman-conf Architecture`, or to use libalpm.

vn971:
Fixed in https://github.com/vn971/rua/commit/c76710a83a916931e44c64375a53eabe2b6ad396  and  https://github.com/vn971/rua/commit/fc548e3f26f006799333675c9ef1ee7358272f26

PR #38, opened 2019-06-13
Use correct arch

Morganamilo:
This was fixed in c76710a8 then later unfixed in fc548e3f.

### DIFF
--- a/src/pacman.rs
+++ b/src/pacman.rs
@@ -4,7 +4,9 @@ use std::collections::HashSet;
 use std::path::Path;
 use std::path::PathBuf;
 use std::process::Command;
+use std::str;
 
+use lazy_static::lazy_static;
 use libalpm::Alpm;
 
 pub fn is_package_installed(alpm: &Alpm, name: &str) -> bool {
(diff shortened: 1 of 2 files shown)


vn971:
Do you have any evidence that `libalpm`-s "arch" is wrong?

vn971:
It uses just the OS architecture instead of things defined in pacman.conf, right?

Morganamilo:
See #9.

https://github.com/vn971/alpm/blob/fork/libalpm/src/util.rs#L9 It's just a wrapper around uname and not related to libalpm at all. I don't really get why it's in there to be honest.

vn971:
Yeah was just about to check the same. Thanks! Merging.

vn971:
Merged. Not making a new release yet as I'm on a vacation with bad internet -- will do later.

Morganamilo:
Also speaking of libalpm. I have my own more complete bindings here: https://github.com/Morganamilo/alpm.rs.

Which I make try and push here some time at some point.

Morganamilo:
Although I do plan use it to write a depsolver first. So probably after then.

vn971:
Interesting! That's a lot of work though. But I don't know how to write bindings and how good it is to fork/rewrite, so I'll just monitor the situation for now.

I'm pretty open to moving to alpm.rs if there'll be any benefits though :+1:

Morganamilo:
The current bindings are not fully complete and have `unimplemented!()` all over. Also as noted in the readme, the bitfields enums are not properly handled, so it's impossible to do a lot of things right.

I originally started trying to make a depsolver a while ago using these bindings. Had issues and looked into patching them, then decided to just start over.

I've already patched rua to use my bindings, and it works as is apart from some function name changes.

But I'm not gonna make a PR for it now at least. If I can make a good depsolver and want to integrate it into rua, then rua would need to use the same alpm bindings as my depsolver.

PR #39, opened 2019-06-19
support multiple search arguments

vn971:


src/action_jailbuild.rs
```diff
@@ -0,0 +1,23 @@
+use crate::rua_dirs::TARGET_SUBDIR;
+use crate::{tar_check, wrapped};
+use directories::ProjectDirs;
+use std::fs;
+use std::path::PathBuf;
+
+pub fn action_jailbuild(offline: bool, target: PathBuf, dirs: &ProjectDirs) {
+	let target_str = target
```
vn971: this code is copy-paste from main.rs, no changes.

Morganamilo: Then why not have it in its own commit?

src/action_search.rs
```diff
@@ -0,0 +1,22 @@
+use crate::print_package_table;
+use raur::{Package, SearchBy};
+
+fn is_package_ok(pkg: &Package, filter: &str) -> bool {
+	pkg.name.contains(filter) || pkg.description.iter().any(|descr| descr.contains(filter))
```
Morganamilo: This should be a case insensitive comparison.

src/action_jailbuild.rs
vn971: I guess I could do that, didn't think of such a desire. Do you review changes per-commit?

If you want, I can split into 2 commits and force-push.

src/action_search.rs
vn971: awesome! Thanks.
Fixed in latest commit.

src/action_jailbuild.rs
Morganamilo: I'm not super anal about other people's repos but I do tend to subscribe to the "one working change" per commit ideology.

>Do you review changes per-commit?

Depends, in this case I could see that commit one is just refactoring so there's no need to read over it.

vn971: Done. I've split it into 2 commits and pushed.

vn971: OK to merge / did you finish review?

src/action_search.rs
Morganamilo: For sake of a clean history. The last commit can easily be squashed into the first via fixup. `git rebase -i HEAD^^^`

vn971: Done! All good now?:)

Just in case: I don't mind rebasing, squashing etc personally. But it's not an official requirement for RUA pull-requestion as well, at least for now. I'm fine with just small commits and readable commit messages from others, and I'll do what others ask if it's not too effortful (rebasing and fixup-ing is totally OK).

src/action_jailbuild.rs
Morganamilo: It looks fine to me. But I'm just casually reading the code. I have not tried to run it.

Although if you wan't a slight optimisation, sort the search terms by size first.

PR #40, opened 2019-06-24
search optimization

vn971:
sort by longest term first, filter by the rest


Morganamilo:
Looks fine to me, not sure why you need review on something so simple.

vn971:
@Morganamilo no real need, just following some procedure. Also, letting CI double-check the branch before merging.

Issue #41, opened 2019-07-01
feature request: offer to run shellcheck on PKGBUILD


vn971:
Offering to run `shellcheck` on PKGBUILD-s can help reviewing. The problem is that a pure invocation of `shellcheck PKGBUILD` will give numerous warnings:
* declared in PKGBUILD but "unused" variables like `$pkgname`
* use of undeclared variables `$srcdir`, `$pkgdir`
* maybe something else

These would better be handled somehow smartly. One approach that I can think of is to create a wrapper "script" which declares variables like `$srcdir`, then sources PKGBUILD and then uses the declared variables like `$pkgname` somehow. Maybe there are other solutions as well.

vn971:
Numerous "unused" warnings have to be handled smartly (`$pkgname` and those kind of variables).

One way, for example, would be to create a wrapper "script" which sources the PKGBUILD and makes use of sourced variables. There could be better solutions, don't take this one blindly.

vn971:
Working on this, hopefully will be done soon.

Issue #1, opened 2018-10-30
implement upstream AUR diff


vn971:
When installing new versions of an already installed package, the user should be able to review changes made upstream. This includes PKGBUILD and other files.

Morganamilo:
I just pushed this crate https://docs.rs/aur-fetch/0.1.0/aur_fetch/ which may be useful to you. Better error messages is still a todo for me but it should be usable.

One slight problem is that I designed it for batch downloads. As in the helper should resolve all the deps via the RPC ahead of time, then pass all of them to this crate and download them at once. Of course doing it in multiple batches is possible but probably makes it much more annoying to use.

And here's an example binary using this crate: https://asciinema.org/a/MnMRusp1XSRo3ZR5MG7PYqlBh

vn971:
@Morganamilo interesting. I must note that it's a hard problem though: the ways how you can combine diffing, fetching and asking user questions is pretty broad. I'll take a look at your proposal.

Morganamilo:
I built this with that in mind. It's a proper library, It never prints anything on it's own. That's up to whoever uses the crate to decide. It does support diffs both as just strings and 'views'.

The view method is pretty much a clone of how aurutils does it if you're familiar with that.

Also this isn't a proposal (otherwise this would be a PR :P) and more hey look at this. But seems as you do want diffs I'd be interested in feedback and if the crate is insufficient in any way for what you want to do.

PR #43, opened 2019-08-16
implement upstream diffs, implement shellcheck

vn971:
* stop ever relying on current dir, current environment etc. This opens possibility for future concurrent workflow.

* add logging

* stop enforcing of `target/` directory being the PKGDEST.
Use more flexible by-file tar review.

* resolve dependencies via RAUR library instead of relying on .SRCINFO-s

* add a separate `shellcheck` command for RUA, with built-in handling of PKGBUILD variables

* move all `include_bytes!` macros to one place

fixes GH-1, fixes GH-41, replaces GH-37.


vn971:
@Morganamilo will you have some time to review the changes in the next days?

Issue #53, opened 2019-08-25
introduce compile-time options for alpm: "shelling" or "linking"


vn971:
Currently, rua always uses the linked library alpm.rs (with underlying bindings to alpm/pacman) to determine installed packages, installable packages, packages outside of main repos, etc.

This is not always desireable as it will break with each ".so" change of the underlying native library. Futhermore, it breaks when switching from `pacman` to `pacman-git` and back. Yet we only use very basic information from alpm -- something we can easily get by calling `pacman` as well.

The downside of calling pacman to obtain information is a bit of speed and unnecessary resource consumption.

The proposal is to have two compile-time options to chose one way of doing it or another. In Rust terms, these are crate features.

vn971:
Closing along with the underlying PR above. This approach is deemed to have too high of a maintenance cost. Regardless whether it's shelling to pacman or linking to libalpm, it's better to only have one way of doing the thing for now.

PR #54, opened 2019-08-25
Support pacman-git

Morganamilo:


vn971:
Should there also be some documentation, so that people can discover this functionality?

Do you want it as split package options, or some comments in the PKGBUILD for manual edits, or just README.md here on github/gitlab ?

Morganamilo:
Could document it the readme. Could also automate it with a check in the pkgbuild.

vn971:
In case of PKGBUILD, how would you do that? Do you think of run-time detection of pacman version and writing corresponding information in `conflicts` ? Or making rua a split package that provides both versions? (Or 3 versions, if issue GH-53 gets solved and we'll have a version without alpm at all.)

I'm thinking of not splitting it for now, but add comments on manual edits that can be done. Curious on your thoughts here. Ah, and a README.md update for sure.

vn971:
If I compile with `--features git`, the code still works with the normal `pacman` quite happily. Does the soname change of pacman actually affect RUA in any way? RUA only uses `find_satisfier` and registers/accesses local/sync db-s I think..

It might continue working just out of pure luck as well -- I don't know. I'm just trying to figure that out.

vn971:
UPD ^

Morganamilo:
try this ;) 

```diff
diff --git a/src/action_install.rs b/src/action_install.rs
index c79547a..a8634fc 100644
--- a/src/action_install.rs
+++ b/src/action_install.rs
@@ -21,6 +21,8 @@ use std::path::PathBuf;
 
 pub fn install(targets: &[String], dirs: &ProjectDirs, is_offline: bool, asdeps: bool) {
        let alpm = pacman::create_alpm();
+    let pkg = alpm.localdb().pkgs().unwrap().find_satisfier("pacman").unwrap();
+    pkg.sync_new_version(None.unwrap());
        let (split_to_raur, pacman_deps, split_to_depth) =
                aur_rpc_utils::recursive_info(targets, &alpm).unwrap_or_else(|err| {
                        panic!("Failed to fetch info from AUR, {}", err);
```

Then `cargo run --featues git`.

The problem is that shared libraries don't really contain any information on types. So what a function returns, arguments it takes and the size of all structs is unknowable. So when you link, you just hope for the best. And if everything is not correctly lined up your program will segfault or worse.

The way libraries get this information is via header files. And as rust can not read c headers instead you must write `-sys` creates. Which are effectively a giant block of `extern "C" {`. Which effectively means: "this c function exists, it returns this type and takes these arguments. trust me".

So if you use the `git` feature with regular pacman then it will compile and run. But not guarantee it will work.

Although it probably will because not that much has changed, the only thing that will cause problems is the error enum having two new variants. Even though you don't match against the errors. If you unwrap one of the new variants, rust normally prints out what was unwrapped so I'm not quite sure what it will do.

One piece of information `.so`s do actually contain though is function names. So if you do apply the above patch the program will then fail to compile because you're trying to call a function that does not exist (it was renamed in the c code) and linking with fail.

Morganamilo:
Also for the pkgbuild:

```sh
if pacman -T pacman-git > /dev/null; then
    _features=git
fi
```

Morganamilo:
Also about the `.so` name changing. It can only change if you're upgrading/downgrading pacman. Code will try to link to whatever `.so` currently exists on disk. So that error does not apply to this situation.

vn971:
> Code will try to link to whatever .so currently exists on disk.

That is true of course. But it's also apparently true that, because alpm didn't change some parts (even though it could), the actual current version _might_ and probably _does_ right now work with `pacman-git`. Nothing to be relied on though. And it will break with soname change for sure.

I'll merge this in, update README and make corresponding changes to PKGBUILD.

PR #55, opened 2019-08-31
WIP: implement "shelling/linking" compilation options

vn971:
The PR should actually compile and work, but I'm not sure by now that the change is really desired.

It introduces a compilation flag(s) that we'll have to maintain in the future, and it makes the code more ugly / less straightforward. Therefore, I'll probably keep this PR for a few weeks, then decide that it's indeed worthless and close it.

fixes GH-53

### DIFF
--- a/src/alpm_shelling.rs
+++ b/src/alpm_shelling.rs
@@ -20,7 +20,7 @@ impl AlpmWrapper for AlpmImpl {
 			.success()
 	}
 
-	fn is_package_installable(&self, package: &str) -> bool {
+	fn is_installed(&self, package: &str) -> bool {
 		Command::new("pacman")
 			.arg("-Sddp")
 			.arg(&package)
--- a/src/alpm_shelling.rs
+++ b/src/alpm_shelling.rs
@@ -1,0 +1,62 @@
+use crate::alpm_wrapper::AlpmWrapper;
+use itertools::Itertools;
+use std::process::{Command, Stdio};
+
+pub struct AlpmImpl {}
+
+pub fn new() -> AlpmImpl {
+	AlpmImpl {}
+}
+
+impl AlpmWrapper for AlpmImpl {
+	fn is_package_installed(&self, package: &str) -> bool {
+		Command::new("pacman")
+			.arg("-T")
+			.arg(&package)
+			.stdout(Stdio::null())
+			.stderr(Stdio::null())
+			.status()
+			.expect("Failed to execute pacman")
+			.success()
+	}
+
+	fn is_package_installable(&self, package: &str) -> bool {
+		Command::new("pacman")
+			.arg("-Sddp")
+			.arg(&package)
+			.stdout(Stdio::null())
+			.stderr(Stdio::null())
+			.status()
+			.expect("Failed to execute pacman")
+			.success()
+	}
+
+	fn list_foreign_packages(&self) -> Vec<String> {
+		let output = Command::new("pacman")
+			.arg("-Qqm")
+			.output()
+			.expect("Failed to execute pacman");
+		if !output.status.success() {
+			eprintln!("Failed to get foreign packages via pacman -Qqm");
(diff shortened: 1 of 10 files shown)


Morganamilo:
My opinion is pick one and stick with it. Maintaining two ways to do the same thing seems kinda pointless. 

Also If I scared you with the .so name changing then do note that it doesn't happen often. Once every major pacman release at most. It's only happened to yay once so far. And the fix is to update your alpm.rs dependency (if needed, it may not be) and then rebuild the package.

src/alpm_shelling.rs
```diff
@@ -0,0 +1,62 @@
...
+use std::process::{Command, Stdio};
+
+pub struct AlpmImpl {}
+
+pub fn new() -> AlpmImpl {
+	AlpmImpl {}
+}
+
+impl AlpmWrapper for AlpmImpl {
+	fn is_package_installed(&self, package: &str) -> bool {
```
Morganamilo: -T includes provides. So this probably does not do what you intend for it to.

vn971: on the contrary, if something already provides a package in the system, I don't need to install it anew?

Morganamilo: That may be true, but the function name is still misleading. If this is intended behaviour then it should be renamed to `find_satisfier` or something.

vn971: You're right about that. I've changed the name to `is_installed` now. This whole branch will probably be discarded eventually though.

vn971:
For the record, when RUA is built with shelling on this branch, this would be the output of `ldd target/debug/rua`:

```
	linux-vdso.so.1 (0x000068e1d8720000)
	liblzma.so.5 => /usr/lib/liblzma.so.5 (0x000068e1d730c000)
	libssl.so.1.1 => /usr/lib/libssl.so.1.1 (0x000068e1d727c000)
	libcrypto.so.1.1 => /usr/lib/libcrypto.so.1.1 (0x000068e1d6fa8000)
	libdl.so.2 => /usr/lib/libdl.so.2 (0x000068e1d6fa3000)
	librt.so.1 => /usr/lib/librt.so.1 (0x000068e1d6f98000)
	libpthread.so.0 => /usr/lib/libpthread.so.0 (0x000068e1d6f77000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x000068e1d6f5b000)
	libc.so.6 => /usr/lib/libc.so.6 (0x000068e1d6d98000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x000068e1d8721000)
	libm.so.6 => /usr/lib/libm.so.6 (0x000068e1d6c52000)
```

vn971:
This is how ldd looks if we further move raur from using openssl (and instead use rustls):
```
	linux-vdso.so.1 (0x00006f0a2aa8d000)
	liblzma.so.5 => /usr/lib/liblzma.so.5 (0x00006f0a293f8000)
	libdl.so.2 => /usr/lib/libdl.so.2 (0x00006f0a293f3000)
	librt.so.1 => /usr/lib/librt.so.1 (0x00006f0a293e8000)
	libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00006f0a293c7000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00006f0a293ad000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00006f0a291ea000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00006f0a2aa8e000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00006f0a290a2000)
```

vn971:
As said in the first message of the PR, let's stick with the current approach of linking to libalpm (alpm.rs) for now.

Issue #59, opened 2019-09-05
make `upgrade` action faster by grouping raur requests


vn971:
subj. Currently, all raur requests in `upgrade` action are done sequentially. Grouping them should improve the speed of `upgrade` action, which is nice, and also reduces load on AUR servers.

Issue #48, opened 2019-08-20
flatten dependency grouping


vn971:
Currently, RUA schedules all dependencies of a package to be build strongly before the main package is built. It thus acts as if all dependencies would be "makedepends". If, however, there are few "makedepends" and many "depends", the build queue can be flattened more aggressively. It will thus further reduce the number of distractions that RUA makes for the user, which is an official value for the project.

PR #63, opened 2019-09-09
MVP: depend on aur-depends

Morganamilo:
This integrates aur-depends into rua.

Hovever this is only a MVP, and aur-depends has been inserted into the
middle of the existing system. There's probably more that could be done
to reduce the code count futher, by removing thing's aur-depends can
already handle.

But I'll leave that for after you decide if you want to include this or
not.

This also ends up fixing #59.


src/action_install.rs
```diff
@@ -81,85 +86,62 @@ fn show_install_summary(pacman_deps: &IndexSet<String>, aur_packages: &IndexMap<
...
-			std::fs::create_dir_all(&build_dir).expect("Failed to create build dir");
-			fs_extra::copy_items(
-				&vec![review_dir],
-				rua_files::global_build_dir(dirs),
-				&CopyOptions::new(),
+	trace!("All expected split packages: {:?}", archive_whitelist);
+
+	for base in &actions.build {
+		let pkgbase = &base.pkgbase;
+		let review_dir = rua_files::review_dir(dirs, pkgbase);
```
vn971: Where is dependency grouping by depth now?
If we don't do that, then the user would be presented with "AUR repo review" for each package separately, with long pauses inbetween where package is actually built (see line 109 `wrapped::build_directory` in new code). If we go with this path, we have to give up the "minimizes user interactions" claim in README.md.

I don't want to do the latter much though, as you'd imagine, because I even wrote specific code myself to minimize it as much as theoretically possible. Even more agressive than grouping deps of individual AUR packages during recursion.

src/action_install.rs
```diff
@@ -1,77 +1,82 @@
...
+		.resolve_targets(targets)
+		.unwrap_or_else(|e| panic!("failed to resolver deps {}", e));
+
+	if !actions.missing.is_empty() {
+		eprintln!("Need to install packages: but they are not found:");
+
+		for Missing { stack, dep } in actions.missing {
+			eprint!("{}", dep);
+			if !stack.is_empty() {
+				eprint!(" (Wanted by: {})", stack.join(" -> "));
```
vn971: One other possible improvement that I had in mind is to let the user see the whole dependency tree during "installation overview" (this function here).
Potential nice things about it: show all unresolved things at once, is useful even if you have no errors, and you can meaningfully highlight different things with color (like dependency recursion depth).

src/action_install.rs
```diff
@@ -81,85 +86,62 @@ fn show_install_summary(pacman_deps: &IndexSet<String>, aur_packages: &IndexMap<
...
+			));
 		}
-		pacman::ensure_aur_packages_installed(files_to_install, asdeps || depth > 0);
 	}
+
+	pacman::ensure_aur_packages_installed(files_to_install, true || as_deps); //TODO hande as_deps properly
 }
 
-pub fn check_tars_and_move(name: &str, dirs: &ProjectDirs, archive_whitelist: &IndexSet<&str>) {
+pub fn check_tars_and_move(name: &str, dirs: &ProjectDirs, archive_whitelist: &[&str]) {
```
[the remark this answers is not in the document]
Morganamilo: This part of the whole "aur-depends has been inserted into the
middle of the existing system. There's probably more that could be done
to reduce the code count futher, by removing thing's aur-depends can
already handle." point.

I don't actually see a reason for this variable to exist at all. If I read correctly, it's a set of all package names? Instead you can just get the package names for an individual base at a time. Which is a much smaller list.

src/action_install.rs
```diff
@@ -1,77 +1,82 @@
...
+pub fn install(
+	resolver: Resolver,
+	targets: &[String],
+	dirs: &ProjectDirs,
+	is_offline: bool,
+	as_deps: bool,
+) {
+	let actions = resolver
+		.resolve_targets(targets)
+		.unwrap_or_else(|e| panic!("failed to resolver deps {}", e));
```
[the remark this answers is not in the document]
Morganamilo: Yep, auto code completion got me.

src/action_install.rs
Morganamilo: I'm not sure how that would happen. This loop is all about reviewing. After the loop `install_all` is called.

So everything is reviewed first, then everything is build.

Morganamilo:
>The real question is: what is the advantage of using the library?

The same reason as libraries I Guess. So programs can be separated into clear distinct parts with a clear API. Programs can re use code and not have to re implement it all the time. I'm not writing all these libraries specifically for rua. I have plans for them and also try to keep them general purpose.

There's also the reason that this library is more correct and has a ton more features of your code.

For example I don't believe your system handles versioned depends. It can't detect conflicts. And it does not handle as deps/explicit properly. It also doesn't consider check_depends.

src/action_install.rs
vn971: Regarding separate loops: sorry, my own naming `rua_files::review_dir` confused me a bit. It can be read as a verb and looks very much alike `reviewing::review_repo`.

On the depth grouping though. You need it because you want to `tar_check` packages in a bulk. Or from another perspective, because the current PR does not have any grouping at all, you **do not install built tar archives until all packages are built**. This way, `makedepends` is not satisfied. Package installation with AUR `makedepends` will break.

vn971: EDIT ^

src/action_install.rs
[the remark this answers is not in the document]
vn971: @Morganamilo yes, that's one possible improvement (not a big one though as IndexSet/HashSet is very fast on `.contains` method regardless).

An even better improvement would be to invoke `makepkg --packagelist` or `makepkg --printsrcinfo`. I think we should strive for the latter. But still, I see no benefits of changing an IndexSet to array here..

src/action_install.rs
Morganamilo: I guess you are right, in the sense that I tested it and it does fail because dependencies are not already installed. But I still don't really know what grouping is, or what tar_check really does.

I will however note that in the current system, it still seems to fail for the same reason. See pacaur for example.

```
morganamilo@Octavia ~ % rua install pacaur
Package pacaur depends on auracle-git. Resolving...

In order to install all targets, the following pacman packages will need to be installed:

And the following AUR packages will need to be built and installed:
  pacaur
  auracle-git

Proceed? [O]=ok, Ctrl-C=abort. o
Reviewing "/home/morganamilo/.config/rua/pkg/pacaur". 
[S]=run shellcheck on PKGBUILD, [D]=(identical to upstream, empty diff), [T]=run shell to edit/inspect, [O]=ok, use package m
Reviewing "/home/morganamilo/.config/rua/pkg/pacaur". 
[S]=run shellcheck on PKGBUILD, [D]=(identical to upstream, empty diff), [T]=run shell to edit/inspect, [O]=ok, use package o
Reviewing "/home/morganamilo/.config/rua/pkg/auracle-git". 
[S]=run shellcheck on PKGBUILD, [D]=(identical to upstream, empty diff), [T]=run shell to edit/inspect, [O]=ok, use package m
Reviewing "/home/morganamilo/.config/rua/pkg/auracle-git". 
[S]=run shellcheck on PKGBUILD, [D]=(identical to upstream, empty diff), [T]=run shell to edit/inspect, [O]=ok, use package o
==> Making package: pacaur 4.8.6-1 (Tue 10 Sep 2019 09:56:36 BST)
==> Checking runtime dependencies...
==> Missing dependencies:
  -> auracle-git
==> Checking buildtime dependencies...
==> ERROR: Could not resolve all dependencies.
Build failed with exit code 8 in /home/morganamilo/.cache/rua/build/pacaur
```

src/action_install.rs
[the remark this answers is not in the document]
Morganamilo: Because hashmap like structures are not magically faster. They have overhead and will be slower unless the array is sufficiently large. And given that most packages are not split packages, I estimate the average size of this array to be 1.

vn971:
Yes, if the library would be correct more often at handling edge cases, and no simple fixes would be possible for the current code base, it would be very appealing.

Most importantly though, from a software design perspective, I'd need it to "be a good abstraction layer". If abstraction has too many details on one of its usages, e.g. doesn't seem to be very re-usable, it's not a good abstraction. For example, I'm quite happy with `raur` and `srcinfo` dependencies -- they're purely awesome. I did chose to not use the "review framework" kind of library in the past though..

So with the current change, I'm still thinking if it's indeed a good separation of concerns. I'll have to verify correctness claims.

vn971:
> And it does not handle as deps/explicit properly.
Can you please raise an issue or expand on that?

> It can't detect conflicts.
A bug needs to be raised for that. I'll try to do it in the evening if you won't do it first (feel free to do it).

> It also doesn't consider check_depends.
fixed in commit 963f59961a6021799ee41eb03fa8ae709a295da3

Morganamilo:
> Can you please expand/explain on that, or raise an issue?

Done. Do note aur-depends already handles all this. It's just not plugged into rua yet because all of your functions apply asdeps to the whole pkgbase and as this is just an MVP I did not change them.

src/action_install.rs
vn971: Ah, interesting. That's a regression due to this issue: https://github.com/vn971/rua/issues/48

`makepkg` enforces not only makedepends to be installed when building, but also just `depends`. Since RUA will enforce makedepends to be present itself, this mechanism should be shut down via `makepkg --nodeps`. Too bad there's no separate option `makepkg --nomakedeps` in addition to `makepkg --nodeps`. I could then use the one that only disables direct dependencies.

I'll fix it today after work, just need to pass in `--nodeps`. The grouping and installation should be valid though (subject to testing after the fix).

Morganamilo: >makepkg enforces not only makedepends to be installed when building, but also just depends

The definition of depends are "things needed at runtime AND build time" while makedepends is "build time only". Passing -d is not really a fix because that will break pkgs that need the depends.

For example a package may call a command outside of any any functions. Thing setting some global variable  `_foo = $(command)`.

And also any git packages will fail to download their sources if git is not installed.

vn971:
@Morganamilo no, I don't see it being merged in the near future TBH, sorry.

1. Additional complexity. I've tried to understand the code behind aur-depends. It seems pretty big, with also yet other dependencies on raur-ext and alpm-utils. All it for something that currently in `aur_rpc_utils.rs`, with the main method being 55LoC, though it doesn't handle versions and maybe never will.

2. I don't see an obvious reason to make the switch. No tempting reason that I could understand, yet.

3. I do not entirely understand what the library does exactly. Raur is a decent building block already. Add some methods that transform the result into a "Map", and it's as good as it gets, to my tastes.

4. The way the library resolves dependencies differs a lot as well. For example, you mentioned that the library might look into AUR to find a version that is not satisfied in main repo, though it does have a package with same name in the main repo (but different version). It goes against how I'd like it to be. I want to never check AUR if there's anything in the repo that satisfies even a different version of what I want.

5. Finally, implementation detail, the current code seems to not group raur information requests a lot. It is, however, done in the current code base. Not really super-important and I'd prefer a correctly working instead of a "fast" one, but since the aim is to have good resolvers in both cases, I tend to like the faster algo.

Also, I'd eventually want to also print the dependency tree the way I described in comments of this issue, and if/when I'll do that, the tree structure will supersede the current improvement of having "stack traces" for unresolved dependencies I think. (Also, even before a tree overview gets implemented, the tool prints resolution logs already. If it goes bad, you still have some material to investigate, should you not want to just drop package installation.)

I'm open to re-considering if I get aware of any actual **bugs** in the code that are not trivial to fix.

Morganamilo:
I'm fine with you closing this but for the record it does group requests as much as possible as well as being faster than your implementation.

vn971:
Ah, indeed, I take back my claim on grouping. Cannot prove or disproof the claim on being faster/slower as well.

PR #66, opened 2019-09-11
Implement naive devel upgrades

gustafla:
This is a feature I really need at least in a basic form. My implementation just considers all packages named `*-git` outdated during upgrade if a `--devel` -flag is passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ RUA will thus interrupt you 3 times, not 7 as if it would be plainly recursive. 
 * This tool focuses on AUR packages only, you cannot `-Suy` your system with it. Use pacman for that.
 * Optional dependencies (optdepends) are not installed. They are skipped. Check them out manually when you review PKGBUILD.
 * The tool does not handle versions. It will always install the latest version possible, and it will always assume that latest version is enough.
-* "-git" packages do not receive automatic upgrades yet. Merge requests welcomed.
+* Development packages such as "-git" packages are only rebuilt when running `rua upgrade --devel`. No version checks are done to avoid unnecessary rebuilds. Merge requests welcomed.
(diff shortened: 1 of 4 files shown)


Morganamilo:
`-git` is not the only type of devel package.

>This is a feature I really need at least in a basic form. My implementation just considers all packages named *-git outdated during upgrade if a --devel -flag is passed.

An improved way is to check if the generated pkgver is newer than the installed version. Then you don't have to rebuild the whole package if there is not update.

gustafla:
Okay, I'll look into that tomorrow. Thanks for help.

vn971:
For a _naive_ implementation, checking package names seems indeed straightforward. E.g. suffixes `-git`, `-hg`, `-bzr`, `-svn`, `-cvs` (minimum, maybe more). Also, all packages where your currently installed version is _higher_ than AUR RPC version.

It is, however, not complete, as you might have a package with a non-standard name. And you could have a package that you've installed exactly at the same time when it was last published. For example, if the maintainer (or you as maintainer) makes an AUR change, you build your package right after that, and the locally built package has thus the same (not newer) version.

Downloading all `git` sources during `upgrade` action is too expensive as well though.. There were some solutions that can "look" inside PKGBUILD-s / SRCINFO-s, find git references, resolve them and check versions, but I don't remember them off the top of my head and it's a bit late for me now.

gustafla:
Thanks for reviewing this. I've fixed the issues you've noted. I hope I didn't miss any. `--devel` is indeed a very common name for the flag in AUR helpers and that's why I like it. The structopt `long` thing turns the field into a long option flag instead of a positional argument. I might look into VCS package version checks next week.

vn971:
@gustafla thanks for the PR! Merged. Only thing I've did is to make the `long` usage uniform with the code around, added a `short` flag, and make some changes around in the old code.

PR #74, opened 2019-09-19
use srcinfo information to improve "builddir" action

vn971:
use srcinfo information to improve "builddir" action
    
also, add support for --force key of makepkg

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ RUA will thus interrupt you 3 times, not 7 as if it would be plainly recursive. 
 * The tool does not handle versions. It will always install the latest version possible, and it will always assume that latest version is enough.
 * Development packages such as "-git" packages are only rebuilt when running `rua upgrade --devel`. No version checks are done to avoid unnecessary rebuilds. Merge requests welcomed.
 * Unless you explicitly enable it, builds do not share user home (~). This may result in maven/npm/cargo/whatever re-downloading dependencies with each build. If you want to override some of that, take a look at ~/.config/rua/wrap_args.d/ and the parent directory for examples.
+* Shared directories such as makepkg.conf-s "PKGDEST", "BUILDDIR" are not supported. Packages are built in isolation from each other, artifacts are stored in standard locations of this tool. Merge requests for moving built artifacts elsewhere are welcomed.
 
 
 ## Safety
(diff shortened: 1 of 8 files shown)


vn971:
The internal goal is to introduce SRCINFO parsing for newly built packages, allowing to refactor the current `archive_whitelist`.

Morganamilo:
Don't you want to use `--packagelist` for this?

vn971:
@Morganamilo it could be done indeed. But won't we need more information in the future? For example, if we'd like to just re-generate the version of the PKGBASE in the future.

Morganamilo:
`--packagelist` includes the version number too. The advantage of `--packagelist` is it gives you the true path to the package. This allows you to support `PKGDEST` and `PKGEXT`. Both of which I have configured and are annoyingly overridden by rua.

vn971:
@Morganamilo so you have those environment variables set, and rua overrides them? Do you mean to say that this code is broken?: https://github.com/vn971/rua/blob/4c86fb1cd4a62c30c395c1a2fbd6a17de301d62a/src/rua_environment.rs#L56

Morganamilo:
No, I have them configured via `makepkg.conf` so they're never in the environment. But because rua sees them as unset, it sets them itself. Which overrides my `makepkg.conf` as the environment takes priority.

vn971:
Interesting, I wasn't aware of this situation. So we can

* teach the client to read makepkg.conf regarding these values, and give read-write access to this directory. That would mean that it won't be isolated between runs and across packages.

* or we can ignore those variables, but do it explicitly, as a design choice rather than accident.

I'll think tomorrow or on the weekend with a clear mind (now it's late).

Morganamilo:
From the perspective of a user. I don't care if it's on purpose or not. I would like my config to be respected.

vn971:
@Morganamilo until RUA starts to move built artifacts to PKGDIST, I've made the current contract exclicit (and also enforcing it) now. In the last commit..

vn971:
Moving already built packages to PKGDIST wouldn't contradict with isolation though, so that's something that can be done in the future.. As far as I understood it so far, the way to go would be to use makepkg "library" (bash) interface and aquire the correct PKGDIST from there.

PR #80, opened 2019-10-16
support alternative command to sudo

vn971:
Another approach could be to not support it in CLI, to avoid bloating.
It would then only be configurable via an environment variable.
It would also, accidentally, make the code simpler.

### DIFF
--- a/src/action_builddir.rs
+++ b/src/action_builddir.rs
@@ -3,9 +3,9 @@ use crate::rua_environment;
 use crate::rua_files::RuaDirs;
 use crate::tar_check;
 use crate::wrapped;
-use std::path::PathBuf;
+use std::path::Path;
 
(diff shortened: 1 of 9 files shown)


vn971:
See the linked issue above for additional context.

src/action_builddir.rs
```diff
@@ -5,7 +5,7 @@ use crate::tar_check;
 use crate::wrapped;
 use std::path::PathBuf;
 
-pub fn action_builddir(dir: PathBuf, dirs: &RuaDirs, offline: bool, force: bool) {
+pub fn action_builddir(dir: &PathBuf, dirs: &RuaDirs, offline: bool, force: bool) {
```
Morganamilo: Like `String` and `&str` you should use `PathBuf` and `&Path`.

vn971: I sometimes forget, unfortunately. Thanks for reminding! Fixed.

Issue #81, opened 2019-10-17
panic when removing non-empty stale build directory


theotherjimmy:
to reproduce, do something like:
 - Control-c a build midway through
 - Attempt to re-run install, rua will tell you that it will remove the build dir.
 - Be okay with that
 - panic!

Such as:
```
❯ rua install infer
WARNING: your AUR repo is up-to-date.
If you continue, the build directory will be removed and the build will be re-run.
If you don't want that, consider resolving the situation manually,
for example:    rua builddir "/home/jimmy/.cache/rua/build/infer"

Reviewing "/home/jimmy/.config/rua/pkg/infer".
[S]=run shellcheck on PKGBUILD, [D]=(identical to upstream, empty diff), [T]=run shell to edit/inspect, [O]=ok, use package o
thread 'main' panicked at 'Failed to remove old build dir "/home/jimmy/.cache/rua/build/infer", No such file or directory (os error 2)', src/action_install.rs:119:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
                                                                  [ 7s168 | Oct 17 10:53AM ]
❯ rua builddir /home/jimmy/.cache/rua/build/infer
==> ERROR: PKGBUILD does not exist.
Build failed with exit code 4 in /home/jimmy/.cache/rua/build/infer
                                                                  [ 0s088 | Oct 17 10:53AM ]
❯ exa /home/jimmy/.cache/rua/build/
bcc             cargo-flamegraph  cargo-watch     infer              slack-desktop
bpftrace        cargo-fuzz        git-series      leiningen
cargo-crev-git  cargo-pkgbuild    git-series-git  onefetch
cargo-edit      cargo-tree        hexyl-bin       shellcheck-static
                                                                  [ 0s001 | Oct 17 10:53AM ]
❯ exa /home/jimmy/.cache/rua/build/infer
infer  pkg  src
```

aacunningham:
I ran into this issue too, so I did a little digging. I haven't really reviewed the changes in the `rm_rf` package, but this issue is resolved with the latest version of `rm_rf` (0.4.1). I can put a PR up later, though I haven't reviewed the changes other than a personal test on my machine with a similar problem.

vn971:
@aacunningham yes indeed. I'll make a fixing PR soon.

PR #83, opened 2019-10-25
Update rm_rf version to 0.4.1

aacunningham:
Version 0.3.0 of `rm_rf` seems to of had a bug and wouldn't properly delete directories. This caused a bug when trying to empty the build directory in `.cache`. Updating the package seems to address this problem. This should address issue #81

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1137,7 +1137,7 @@ dependencies = [
 
 [[package]]
 name = "rm_rf"
-version = "0.3.0"
+version = "0.4.1"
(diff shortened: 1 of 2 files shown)


vn971:
This was apparently the correct fix: https://github.com/vn971/rua/pull/84/files
Thanks anyway though, your PR would otherwise be correct, if the other library didn't unfortunately change the semantics of the method.

vn971:
In the latest version of the library, `remove` will fail for unexistent path, but `ensure_removed` will siletly ignore the lack of destination path.

PR #88, opened 2019-11-01
Deduplicate info code

aacunningham:
Addresses #73, removing the `raur::info` call at the top of the function and using `aur_rpc_utils::info_map` instead.

This includes a modification to the `info_map` arguments. This change more closely matches the signature of `raur::info` and allows both `&[String]` and `&[&str]` to be passed to it.


vn971:
Thanks! All good, merging.

BTW, if you're thinking on taking something bigger or would like to ask anything interactively, feel free to join #rua on freenode (IRC).
It's completely optional though, just a chat-like thing when needed.

PR #94, opened 2019-11-23
add support for zstd (zst) archive format

vn971:


Morganamilo:
I know you have a thing for rust libs. But libarchive supports pretty much all archive types so you'd only need the one dep. Also pacman deps it so it's not really an extra dep.

vn971:
@Morganamilo you have a point. Especially considering that adding each new type of archive is a manual job right now.
On the other hand, both `zstd` and `xz` had "code execution" CVE-s in the past, as late as 2019 and 2017 respectfully. I really don't know if that should be traded for the downsides of the pure-rust packages.

Personally, I prefer more safety against CVE-s and no soname breakages over feature variety. E.g. `7z` and `rar` are off the table now, and the tool doesn't even attempt to recognize a wrongly named archive (archive format detection).

If I see a lot of people would want the project to move in another direction here, I might change that.

PR #104, opened 2020-02-17
fix #102

### LINKED ISSUE #102: install instructions incorrect (lz)
Line 43 in the README.md fails since there is no `lz` package.
Looking at https://aur.archlinux.org/packages/rua/ it seems to be a typo, intended to mean `xz`.

marcusatiliusregulus:
fixes #102 
by changing non-existent `lz` dependency to `xz`, which is the one listed on the aur page

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ RUA is a build tool for ArchLinux, AUR. Its features:
 
 ## Install dependencies
 ```sh
-sudo pacman -S --needed git base-devel bubblewrap-suid lz shellcheck cargo
+sudo pacman -S --needed git base-devel bubblewrap-suid xz shellcheck cargo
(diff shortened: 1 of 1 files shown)


vn971:
Hi, thanks for the pull request! The `lz` package was certainly no a typo, this package was available in pacman, and needed for RUA to work. Now, however, it seems that `xz` provides both the library and CLI tools, so probably we don't need `lz` or its alternatives anymore. Thanks for raising awareness for that! I need to update the `depends` section on AUR itself as well. Will do shortly.

vn971:
UPD: in the PKGBUILD it is `xz` already, so maybe it is an older mistake after all. Anyway, merged, and thanks again!

PR #109, opened 2020-03-21
Fix a panic when moving tarballs across storage devices

VukoDrakkeinen:
I have `$XDG_CACHE_HOME` and `$XDG_DATA_HOME` on different disks, which would trip up `rua`:
```
Invalid cross-device link (os error 18)', src/action_install.rs:232:13
```

### DIFF
--- a/src/action_install.rs
+++ b/src/action_install.rs
@@ -228,11 +228,27 @@ pub fn check_tars_and_move(name: &str, dirs: &RuaDirs, archive_whitelist: &Index
 	});
 
 	for (file, file_name) in dir_items {
-		fs::rename(&file.path(), checked_tars_dir.join(file_name)).unwrap_or_else(|e| {
-			panic!(
-				"Failed to move {:?} (build artifact) to {:?}, {}",
-				&file, &checked_tars_dir, e,
-			)
-		});
+		let src = &file.path();
+		let dst = &checked_tars_dir.join(file_name);
+
+		fs::rename(src, dst)
+			.or_else(|err| {
+				// EXDEV (invalid cross-device link) gets aggregated into io::ErrorKind::Other
+				if err.raw_os_error() != Some(libc::EXDEV) {
+					return Err(err);
+				}
+
+				// can't move across disks, copy & delete instead
+				fs::copy(src, dst)?;
+				let _ = fs::remove_file(src);
+
+				Ok(())
+			})
+			.unwrap_or_else(|e| {
+				panic!(
+					"Failed to move {:?} (build artifact) to {:?}, {}",
+					&file, &checked_tars_dir, e,
+				)
+			});
 	}
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1231,6 +1231,7 @@ dependencies = [
  "indexmap 1.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "itertools 0.8.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.66 (registry+https://github.com/rust-lang/crates.io-index)",
  "libflate 0.1.27 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "prettytable-rs 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ term_size = "0.3.1"
 uname = "0.1.1"
 users = "0.9.1"
 xz2 = "0.1.6"
+libc = "0.2.66"
 
 [profile.release]
 lto = true


src/action_install.rs
```diff
@@ -228,11 +228,26 @@ pub fn check_tars_and_move(name: &str, dirs: &RuaDirs, archive_whitelist: &Index
...
-				"Failed to move {:?} (build artifact) to {:?}, {}",
-				&file, &checked_tars_dir, e,
-			)
-		});
+		let src = &file.path();
+		let dst = &checked_tars_dir.join(file_name);
+
+		fs::rename(src, dst)
+			.or_else(|err| {
+				const EXDEV: i32 = 18; //Invalid cross-device link
```
vn971: I wonder on two things:
* is there any way to reference this number (`18`) from stdlib or some crate that we already have in imports? Your carefully written PR suggests that you probably tried to look it up, but doesn't hurt asking. If you tried finding it but it was fruitless, then it's OK, just asking. I'll take another glance at it tomorrow.

* are there other reasons where rename would fail, but copying wouldn't? E.g. should we always try to copy as fallback?

vn971: UPD: I'll better take a fresh look tomorrow.

VukoDrakkeinen: _(Err, you seem to have deleted your comment while I was typing a response, but let me address your concerns anyway :stuck_out_tongue:)_

I felt that pulling in the whole [`libc`](https://lib.rs/libc) for a single stable numeric value would be an overkill, but your comment made me reconsider.  
It seems that `rua` depends on `libc` transitively anyway, so I guess we could use [`libc::EXDEV`](https://docs.rs/libc/0.2.68/libc/constant.EXDEV.html) here instead?

src/action_install.rs
```diff
@@ -228,11 +228,26 @@ pub fn check_tars_and_move(name: &str, dirs: &RuaDirs, archive_whitelist: &Index
...
+				if err.raw_os_error() != Some(EXDEV) {
```
VukoDrakkeinen: _(original question was: "should we always fall back to copying?")_

Good point! A cursory glance at `man 3 rename` does not seem to bring up anything, though, and I would like to preserve the original error if possible.  
Hmm, I'll look for other possible cases tomorrow after I wake up.

VukoDrakkeinen: So instead of going to sleep I went through the source code of GNU `mv`... ¯\\\_(ツ)_/¯

`movefile()` calls `do_move()` which, weirdly enough, calls `copy()` with `move_mode = true`.
This attempts to `renameatu()` at first and seems to fall back to copying [ONLY](https://github.com/coreutils/coreutils/blob/9b4bb9d28a6a5f84c407f795d518726fd7902121/src/copy.c#L2466) when `errno == EXDEV`,
erroring out otherwise.
Of note are this [giant disclaimer](https://github.com/coreutils/coreutils/blob/9b4bb9d28a6a5f84c407f795d518726fd7902121/src/copy.c#L2451-L2465) about exotic error codes and this [comment about race conditions](https://github.com/coreutils/coreutils/blob/9b4bb9d28a6a5f84c407f795d518726fd7902121/src/copy.c#L2468-L2477).
I might have _missed_ something, as the code is quite complex.

Things learned: moving files robustly is apparently REALLY complicated.

Fun fact: ZFS can [apparently](https://unix.stackexchange.com/questions/452620/is-mv-atomic-operation-between-two-file-systems/452624#comment822084_452624) `rename()` files across disks as long as they belong to the same zpool. Fancy!

src/action_install.rs
vn971: (sorry about deletion, it was by far past midnight for me, so I felt hesitant whether I'm missing some simple alternative here)

On the libc question. Yes, since we already have `libc` in, I'd say we should use it. Can you change that part please?

src/action_install.rs
vn971: Wow, that's some very impressive research!
// Also, that first line made me laugh, all too familiar...:D

I tried to look up if there's some crate that would catch these juicy topics, but there doesn't seem to be any. Apparently, the code's not big enough to create a crate. I wonder if the situation calls for making a QA stackoverflow post though. That doesn't hold the PR or anything, just thinking.

vn971:
One request to please use the libc import, otherwise LGTM :+1:

src/action_install.rs
VukoDrakkeinen: Terribly sorry for the late response! I've been feeling somewhat under the weather.

_(And no worries, that's perfectly understandable! They were very much valid remarks, though, which why I decided to answer them anyway.)_

Wilco!

src/action_install.rs
VukoDrakkeinen: Thanks, I felt that `rua` deserved it! :)

It's telling that even `coreutils` folks find the situation uncertain.

It might be common enough that it warrants an addition to `std`. `std::io::ErrorKind::MoveAcrossDevices` and `std::fs::move()`..?

vn971:
Thanks a lot for the pull request, merged it! I made the story with coreutils, our considerations over it and our behavior a bit more explicit here: 296482671587f79604db0ec4cc648d3cdefbda78 before merging. Also making a new release now.

PR #110, opened 2020-03-21
Respect makepkg config files in addition to environment variables

VukoDrakkeinen:
Previously, if set only e.g. in `/etc/makepkg.conf`, options
like `PKGEXT` were ignored and defaults were used instead.

Unfortunately, since `makepkg` configuration files are valid
scripts, reading them involves shelling out. Non-ideal, but I
feel that ignoring them is strictly worse.

### DIFF
--- a/res/print_makepkg_config.sh
+++ b/res/print_makepkg_config.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/bash
+
+# config.sh is a publicly exposed utility library:
+# https://git.archlinux.org/pacman.git/commit/scripts/libmakepkg/util/config.sh.in?id=a00615bfdad628299352b94e0f44d211a758fd17
+source "${LIBRARY:-/usr/share/makepkg}/util/config.sh";
+
+load_makepkg_config;
+
+# config entries which can be overriden with environment variables; taken from util/config.sh
+for var in PKGDEST SRCDEST SRCPKGDEST LOGDEST BUILDDIR PKGEXT SRCEXT GPGKEY PACKAGER CARCH; do
+	[[ -v $var ]] && printf "%s=%s\0" "$var" "${!var}";
+done
--- a/src/action_builddir.rs
+++ b/src/action_builddir.rs
@@ -1,12 +1,11 @@
 use crate::pacman;
-use crate::rua_environment;
-use crate::rua_files::RuaDirs;
+use crate::rua_environment::RuaEnv;
 use crate::tar_check;
 use crate::wrapped;
 use std::path::Path;
 use std::path::PathBuf;
 
-pub fn action_builddir(dir: &Option<PathBuf>, dirs: &RuaDirs, offline: bool, force: bool) {
+pub fn action_builddir(dir: &Option<PathBuf>, rua_env: &RuaEnv, offline: bool, force: bool) {
 	// Set `.` as default dir in case no build directory is provided.
 	let dir = match dir {
 		Some(path) => &path,
@@ -18,18 +17,18 @@ pub fn action_builddir(dir: &Option<PathBuf>, dirs: &RuaDirs, offline: bool, for
 	let dir_str = dir
 		.to_str()
 		.unwrap_or_else(|| panic!("{}:{} Cannot parse CLI target directory", file!(), line!()));
-	wrapped::build_directory(dir_str, &dirs, offline, force);
+	wrapped::build_directory(dir_str, &rua_env.dirs, offline, force);
 
-	let srcinfo = wrapped::generate_srcinfo(dir_str, dirs).expect("Failed to obtain SRCINFO");
+	let srcinfo =
+		wrapped::generate_srcinfo(dir_str, &rua_env.dirs).expect("Failed to obtain SRCINFO");
 	let ver = srcinfo.version();
-	let ext = rua_environment::extension();
 	let archive_names = srcinfo.pkgs.iter().map(|package| {
 		let arch = if package.arch.contains(&*pacman::PACMAN_ARCH) {
 			pacman::PACMAN_ARCH.to_string()
 		} else {
 			"any".to_string()
 		};
-		format!("{}-{}-{}{}", package.pkgname, ver, arch, ext)
+		format!("{}-{}-{}{}", package.pkgname, ver, arch, rua_env.pkgext)
 	});
 
 	for archive_name in archive_names {
--- a/src/rua_environment.rs
+++ b/src/rua_environment.rs
@@ -1,11 +1,19 @@
 use crate::cli_args;
 use crate::cli_args::CLIColorType;
 use crate::cli_args::CliArgs;
+use crate::rua_files::RuaDirs;
 use chrono::Utc;
+use colored::Colorize;
 use env_logger::Env;
 use log::debug;
 use std::env;
 use std::io::Write;
+use std::process::Command;
+
+pub struct RuaEnv {
+	pub dirs: RuaDirs,
+	pub pkgext: String,
+}
 
 pub fn set_env_if_not_set(key: &str, value: &str) {
 	if env::var_os(key).is_none() {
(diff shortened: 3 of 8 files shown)


res/print_makepkg_config.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/bash
+
+source "${LIBRARY:-/usr/share/makepkg}/util/config.sh";
```
VukoDrakkeinen: This is apparently public, as per [this `makepkg` commit message](https://git.archlinux.org/pacman.git/commit/scripts/libmakepkg/util/config.sh.in?id=a00615bfdad628299352b94e0f44d211a758fd17)

res/print_makepkg_config.sh
```diff
@@ -0,0 +1,10 @@
...
+
+load_makepkg_config;
+
+# config entries which can be overriden with environment variables; taken from util/config.sh
+for var in PKGDEST SRCDEST SRCPKGDEST LOGDEST BUILDDIR PKGEXT SRCEXT GPGKEY PACKAGER CARCH; do
```
VukoDrakkeinen: The alternative to hard-coding these would be to grep `util/config.sh`,
which would be just as brittle (if not more).

src/action_builddir.rs
```diff
@@ -18,18 +17,18 @@ pub fn action_builddir(dir: &Option<PathBuf>, dirs: &RuaDirs, offline: bool, for
 	let dir_str = dir
 		.to_str()
 		.unwrap_or_else(|| panic!("{}:{} Cannot parse CLI target directory", file!(), line!()));
-	wrapped::build_directory(dir_str, &dirs, offline, force);
+	wrapped::build_directory(dir_str, &rua_env.dirs, offline, force);
```
VukoDrakkeinen: I felt that specifying `rua_env.dirs` everywhere makes code easier
to grok, being more explicit. It makes for a noiser diff, though.

src/rua_environment.rs
```diff
@@ -49,51 +56,56 @@ pub fn prepare_environment(config: &CliArgs) {
...
+		line.find('=')
+			.map(|sep_pos| (&line[..sep_pos], &line[sep_pos + 1..]))
+	});
+
+	// config entries won't appear here unless set
+	for (var, value) in config_entries {
+		debug!("makepkg option: {} = {}", var, value);
+		match var {
+			"PKGDEST" | "SRCDEST" | "SRCPKGDEST" | "LOGDEST" | "BUILDDIR" => {
+				panic!("Cannot work with {} set. Please run RUA without it", var)
```
VukoDrakkeinen: Now that we also read configuration files, should we downgrade this
to a warning, maybe?

src/rua_files.rs
```diff
@@ -21,6 +21,8 @@ pub struct RuaDirs {
 	global_checked_tars_dir: PathBuf,
 	/// Script used to wrap `makepkg` and related commands
 	pub wrapper_bwrap_script: PathBuf,
+	/// Script used to read `makepkg` config
```
VukoDrakkeinen: Now that the struct also contains paths to 2 files, should we
rename it to `RuaPaths`? (e.g. `rua_env.paths.review_dir()`)

res/print_makepkg_config.sh
vn971: Can you please add this as a comment in the script as well, for future reference?

src/rua_environment.rs
vn971: With respect to the wording, it'd be nicer to be more detailed in this `panic!`. Maybe "Cannot work with makepkg configuration {}. Please ..." ?

vn971: Yeah, I guess you're right.
It was very easy to start with the mindset "instantly bail if something might not meet user-s expectations". E.g. instead of working around some difficulties, only really work in the happy scenario. But here, the option to still continue with a warnings seems better suited than completely breaking, if you think about it.

tl&dr; agree, let's change that.

src/rua_environment.rs
```diff
@@ -49,51 +56,56 @@ pub fn prepare_environment(config: &CliArgs) {
...
+
+	let config = Command::new(&dirs.makepkg_config_loader)
+		.output()
+		.unwrap_or_else(|e| panic!("Reading makepkg config failed: {}", e))
+		.stdout;
+	let config = String::from_utf8(config).expect("makepkg config contains non-UTF-8 data");
+
+	// format: `VAR=VALUE\0`
+	let config_entries = config.split_terminator('\0').filter_map(|line| {
+		line.find('=')
```
vn971: Instead of `filter_map`-ing, should we panic when a line doesn't have `=` but is non-empty? E.g. first filter out the non-empty sections, and then for each of them, expect that the `=` should be there.

src/rua_environment.rs
```diff
@@ -49,51 +56,56 @@ pub fn prepare_environment(config: &CliArgs) {
...
+		}
+	}
+
+	for &var in &["PKGDEST", "SRCDEST", "SRCPKGDEST", "LOGDEST", "BUILDDIR"] {
+		env::set_var(var, "/dev/null"); // make sure we override it later
+	}
+
+	RuaEnv {
+		dirs,
+		pkgext: pkgext.expect("Internal error: PKGEXT is unset"),
```
vn971: Can we highlight the fact that this is makepkg-related? E.g. `makepkg config parameter PKGEXT not found`. Or anything more verbose, if seen fit.

vn971:
Besides the minor comments above, everything seems good. Thanks a lot for the PR! I'll also need to playtest it a bit, but afterwards, I expect to merge.

res/print_makepkg_config.sh
VukoDrakkeinen: Good point! :+1:

src/rua_files.rs
VukoDrakkeinen: I would like to do it in a separate PR, if you don't mind.

src/rua_environment.rs
VukoDrakkeinen: Hm, I guess that'd be stricter and, by extension, ever so slightly more likely to catch any anomalies. Alright!

VukoDrakkeinen:
Man, writing concise and clear error messages/warnings is HARD. I'm still not fully satisfied, but I think it should be somewhat fine now. What do you think?

src/rua_files.rs
vn971: Sure, it's totally fine.

src/rua_environment.rs
vn971: Yeah, there's almost no chance that the stricter version will ever catch anything - still, semantically it more correctly represents our expectations.
Thanks!

vn971:
Merged! Thanks a lot!
I did touch up the logging just a tiny bit: https://github.com/vn971/rua/commit/db9c6867fdc5dc82627cda6466196cb1448b5edb  An extra pair of eyes is always useful in things like docs. Also, I removed the "$" from the logging if you don't mind, because this symbol might not necessarily be written by the user, and could possibly confuse (IDK? Just left it out for simplicity).

vn971:
UPD: hmm, we apparently didn't catch it and there were no tests for that, but `rua` now breaks if you execute a command like `rua search abcd` from root. It will complain saying:
```
rua search test
RUA does not allow building as root.
Also, makepkg will not allow you building as root anyway.
```

so basically, the `users::get_current_uid()` check is now done earlier/eagerly, and we construct `RuaPaths` even when we don't need it (as e.g. for `search` action).

I'll think on how to de-compose those two back.

vn971:
I think the problem can be solved by moving the `makepkg` config checks to the function that instantiates RuaPaths, because one only makes sense with the other, currently.
I'll try to do that now.

VukoDrakkeinen:
> UPD: hmm, we apparently didn't catch it and there were no tests for that, but `rua` now breaks if you execute a command like `rua search abcd` from root.

Hmm, in a similar vein, non-building commands (`search`, `info`, etc) will fail when executed inside `rua` (`Error: another RUA instance already running.`)

vn971:
@VukoDrakkeinen yeah, indeed. I'd still tackle that problem separately, it shouldn't be hard.

Meanwhile I'm finishing the fix to make `root` and `PKGEXT` evaluations "lazy" (only for needed commands).

Issue #112, opened 2020-03-27
running "passive" commands like `rua search` fail for root


vn971:
Same error happens if you e.g. define an incompatible PKGEXT environment variable, even though `rua search` never needs to do anything PKGEXT or makepkg-related.

The current error log is:
```
rua search test
RUA does not allow building as root.
Also, makepkg will not allow you building as root anyway.
```

This is a regression from the new feature https://github.com/vn971/rua/pull/110 (released as part of `v0.17.4` yesterday).

PR #113, opened 2020-03-27
do not instantiate RuaPaths for passive commands

vn971:
fixes GH-112

### DIFF
--- a/src/action_upgrade.rs
+++ b/src/action_upgrade.rs
@@ -2,7 +2,7 @@ use crate::action_install;
 use crate::aur_rpc_utils;
 use crate::pacman;
 use crate::print_package_table;
-use crate::rua_environment::RuaEnv;
+use crate::rua_paths;
 use crate::terminal_util;
 use alpm::Version;
 use colored::*;
@@ -23,7 +23,7 @@ fn pkg_is_devel(name: &str) -> bool {
 	RE.is_match(name)
 }
 
-pub fn upgrade(rua_env: &RuaEnv, devel: bool, printonly: bool) {
+pub fn upgrade(devel: bool, printonly: bool) {
 	let alpm = pacman::create_alpm();
 	let pkg_cache = alpm
 		.localdb()
@@ -93,12 +93,13 @@ pub fn upgrade(rua_env: &RuaEnv, devel: bool, printonly: bool) {
 		} else {
 			print_outdated(&outdated, &unexistent);
 			eprintln!();
+			let paths = rua_paths::RuaPaths::new();
 			loop {
 				eprint!("Do you wish to upgrade them? [O]=ok, [X]=exit. ");
 				let string = terminal_util::read_line_lowercase();
 				if &string == "o" {
 					let outdated: Vec<String> = outdated.iter().map(|o| o.0.to_string()).collect();
-					action_install::install(&outdated, &rua_env, false, true);
+					action_install::install(&outdated, &paths, false, true);
 					break;
 				} else if &string == "x" {
 					break;
--- a/src/action_upgrade.rs
+++ b/src/action_upgrade.rs
@@ -2,7 +2,7 @@ use crate::action_install;
 use crate::aur_rpc_utils;
 use crate::pacman;
 use crate::print_package_table;
-use crate::rua_paths;
+use crate::rua_paths::RuaPaths;
 use crate::terminal_util;
 use alpm::Version;
 use colored::*;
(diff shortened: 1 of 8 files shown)


vn971:
@VukoDrakkeinen do you like reviewing things BTW? :) An extra pair of eyes is always better than solo.

src/rua_paths.rs
```diff
@@ -21,13 +25,16 @@ pub struct RuaPaths {
 	global_checked_tars_dir: PathBuf,
 	/// Script used to wrap `makepkg` and related commands
 	pub wrapper_bwrap_script: PathBuf,
-	/// Script used to read `makepkg` config
-	pub makepkg_config_loader: PathBuf,
+	/// makepkg configuration for PKGEXT
+	pub makepkg_pkgext: String,
```
vn971: This is the core of the PR. `PKGEXT` is made part of `RuaPaths`.

src/action_upgrade.rs
```diff
@@ -93,12 +93,13 @@ pub fn upgrade(rua_env: &RuaEnv, devel: bool, printonly: bool) {
 		} else {
 			print_outdated(&outdated, &unexistent);
 			eprintln!();
+			let paths = rua_paths::RuaPaths::new();
```
VukoDrakkeinen: Hm, this is minor, but are we sure we want to check for
other instances conditionally in `rua upgrade`?
This might be surprising to users.

src/rua_paths.rs
VukoDrakkeinen: Oh boy, RIP accurate name for `RuaPaths`, we hardly knew ye :P

Overall the PR looks fine :+1:, but we might want to look into making some
side-effects more obvious in the future, so a similar regression won't happen again.

src/action_upgrade.rs
vn971: Well the other scenario is the `rua upgrade --printonly`. Is it bad to run something in parallel with it?

This subcommand will fetch something from AUR and check the local database, but that's it. It doesn't change state, and it exists immediately (well, once HTTP things are done). Thoughts?

src/rua_paths.rs
vn971: Yeah, I started adding comments to those methods now. Also, changed the method name from `RuaPaths::new()` to `RuaPaths::initialize_paths()` in last commit. It has more "verb" to it I guess.

src/action_upgrade.rs
VukoDrakkeinen: Yeah, I guess you're right. `--printonly` should be fine without the checks.

I think what irks me is that both `action_install()` and `action_builddir()`
take `&RuaPaths` as an argument, but `action_upgrade()` does not
and instead constructs one internally. This hides the fact that the root/instance
checks are performed.

How about something like this?:
```rust
//main.rs:
if (printonly) {
  action::upgrade_printonly(devel);
} else {
  let rua_paths = RuaPaths::initialize_paths();
  action::upgrade(&rua_paths, devel);
}

//action_upgrade.rs:
fn some_common_upgrade_code();

fn upgrade() {
  some_common_upgrade_code();
  //...upgrade-specific code...
}

fn upgrade_printonly() {
  some_common_upgrade_code();
  //..printonly specific code...
}
```

vn971: *sigh* you're right of course. Even though it's a bit messy to actually implement.

I've played around with creating common method and things on top of it, but it was a bit unpleasant. I've then tried to employ an enum to handle the action:
```
enum UpgradeType {
	Printonly,
	Real { paths: RuaPaths },
}
```
which allows things to be a bit more concise. WDYT?

vn971: pushed the changes --^

VukoDrakkeinen: Hm, I think we can do better.

I took a quick stab at this at [draft/rua-lazy-paths](https://github.com/VukoDrakkeinen/rua/commit/17c862fc710606edf87939e2de0cab8e95f70340), and you're right that it's somewhat
messy. There is some minor duplication here and there, and the return types are fugly.

Still, I think that it would be the right way going forward. The enum way feels like a workaround.

VukoDrakkeinen: (Man, I hope I'm not being overly harsh here)

vn971: I'd say the approaches are very similar already, but you were opinionated, and there's no problem for me with going for the latter approach, so merged that one in. Will make a release now so that `rua upgrade --printonly` works from root.

VukoDrakkeinen: Cool, that also works for me :+1:

PR #115, opened 2020-03-27
[DRAFT] Re-factor `tar_check` and SandboxOpts

VukoDrakkeinen:
So, uh, I went a _little bit_ overboard with **reckless refactoring** while fixing the bug ¯\\\_(ツ)_/¯

I'm not sure I want to merge it as is, really, but here it is just in case.
Alternate (simpler!) take: #116

Thoughts?

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,1 +1,2 @@
 hard_tabs = true
+overflow_delimited_expr = true
--- a/src/action_builddir.rs
+++ b/src/action_builddir.rs
@@ -5,22 +5,15 @@ use crate::wrapped;
 use std::path::Path;
 use std::path::PathBuf;
 
-pub fn action_builddir(dir: &Option<PathBuf>, rua_env: &RuaEnv, offline: bool, force: bool) {
+pub fn action_builddir(dir: Option<PathBuf>, rua_env: &RuaEnv, offline: bool, force: bool) {
 	// Set `.` as default dir in case no build directory is provided.
-	let dir = match dir {
-		Some(path) => &path,
-		None => Path::new("."),
-	};
-	let dir = dir
-		.canonicalize()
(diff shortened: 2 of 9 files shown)


VukoDrakkeinen:
I'm going to post some explanatory comments tomorrow,
which should make this PR (hopefully) more digestible.

vn971:
Can you please explain a bit on the underlying issue itself? E.g. you try to do what, and what breaks?

VukoDrakkeinen:
> Can you please explain a bit on the underlying issue itself? E.g. you try to do what, and what breaks?

Ah, sorry, I shouldn't have removed the commit message from the PR itself.

As you might know by now, I have a bit of a unique setup with my `$XDG_CACHE_HOME`.
Building any VCS package with `rua` on my system would always emit a warning:
```
==> WARNING: PKGBUILD is not writeable -- pkgver will not be updated
```
and I would get the stale package version.

I tracked this down to a working dir & bind mounted dir paths mismatch.
Turns out, that `chdir()` internally canonicalises the path, so in the end
`$PWD != $pkgbuild_dir` (in `bwrap --bind $pkbuild_dir $pkgbuild_dir`)
and Bubblewrap blocks write access.

The fix is to eagerly canonicalise the path when passing it to `wrap.sh`.

...and then I got extremely carried away while making minor clean-ups
in `wrapped.rs` :d

VukoDrakkeinen:
Honestly, the more I look at this PR, the less ready it looks. :|
I think I'm going to try to polish it over the coming week, splitting stuff
into separate commits and such.

VukoDrakkeinen:
We might want to think about merging #116 as a hotfix, though.

vn971:
Yes, exactly that:
> splitting stuff
> into separate commits and such

Can you do that? Would be very helpful. I can merge the fix itself though.. Which I'll try to do delaying my sleepiness now.

vn971:
Closing due to inactivity.
Feel free to return to this and re-open, should there be time for this PR later.

PR #116, opened 2020-03-27
Fix symlinks making jailed PKGBUILDs read-only (take B)

VukoDrakkeinen:
A vastly simpler take on #115.

### DIFF
--- a/src/action_install.rs
+++ b/src/action_install.rs
@@ -121,6 +121,9 @@ fn install_all(
 		for (pkgbase, _depth, _split) in &packages {
 			let review_dir = rua_env.dirs.review_dir(pkgbase);
 			let build_dir = rua_env.dirs.build_dir(pkgbase);
+			let build_dir = build_dir
+				.canonicalize()
+				.unwrap_or_else(|e| panic!("Cannot canonicalize path {:?}, {}", build_dir, e));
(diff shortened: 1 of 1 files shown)


VukoDrakkeinen:
(tweaked commit message)

src/action_install.rs
```diff
@@ -121,6 +121,9 @@ fn install_all(
 		for (pkgbase, _depth, _split) in &packages {
 			let review_dir = rua_env.dirs.review_dir(pkgbase);
 			let build_dir = rua_env.dirs.build_dir(pkgbase);
+			let build_dir = build_dir
+				.canonicalize()
+				.unwrap_or_else(|e| panic!("Cannot canonicalize path {:?}, {}", build_dir, e));
```
vn971: "Cannot canonicalize build_dir ..." ? E.g. being a bit more specific on where the canonicalization even happens?
Also, including the exact pkgbase would help - just in case?

vn971: Although I better just do it myself, as easy as re-fetch everything, merge etc.

vn971:
Merged. Thanks for the PR and the detailed description!

VukoDrakkeinen:
Happy to help! :)

PR #122, opened 2020-04-30
Add --ignore option to upgrade action

### LINKED ISSUE #107: One-time ignoring package while upgrading
Subj. It would be cool if `rua upgrade` has option to select which packages to ignore in this run.

mqus:
Fixes #107 by adding a new option to the upgrade command.

The following usages mean the same thing:
```
rua upgrade --ignore a --ignore b
rua upgrade -I a -I b
rua upgrade --ignore a b
rua upgrade -I a b
```
I took `-I` and not `-i` because `-i` already has another meaning in the pacman command (show Information) and this opens up rua to use the same letter for the same feature in the future.

But of course this is only my opinion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ RUA is a build tool for ArchLinux, AUR. Its features:
 
 `rua install xcalib`  # install or upgrade a package
 
-`rua upgrade`  # upgrade all AUR packages. You can selectively ignore packages by adding them to `IgnorePkg` in `pacman.conf` (same as with non-AUR packages and `pacman`). You can upgrade only specific packages with `rua install A B C`.
+`rua upgrade`  # upgrade all AUR packages. You can selectively ignore packages by using `--ignore` or adding them to `IgnorePkg` in `pacman.conf` (same as with non-AUR packages and `pacman`). You can upgrade only specific packages with `rua install A B C`.
 
 `rua shellcheck path/to/my/PKGBUILD`  # run `shellcheck` on a PKGBUILD, discovering potential problems with the build instruction. Takes care of PKGBUILD-specific variables.
 
--- a/src/action_upgrade.rs
+++ b/src/action_upgrade.rs
@@ -22,9 +22,9 @@ fn pkg_is_devel(name: &str) -> bool {
 	RE.is_match(name)
 }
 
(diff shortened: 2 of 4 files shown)


vn971:
Thanks for the PR!

Speaking about pacman "style" compatibility. I think capital letters always denote the "command" in pacman? E.g. `-D`, `-S`, `-F` etc are _actions_, but the rest are lowercase _options_? From that regard, `--ignore` should not have a short version, or use a different letter? What do you think?

mqus:
Ah right, I forgot that. I would really like a short version but if i only have to type `--ignore` once for it to work with multiple packages then I won't mind only having the long version.

vn971:
Yeah, actually I wanted to say that as well (forgot). Since package names are guaranteed to have no comma `,` ,what do you think about specifying a comma-separated list of packages to ignore?

vn971:
E.g. `--ignore a,b,c --ignore d,e`  would be "equal" to `--ignore a,b,c,d,e`

mqus:
Well currently I rely on structops which does the same but with a space instead of a comma (like written in the PR description). I don't really see a need to change this. My previous comment could be misunderstood so I'll repeat it in other words: I'm ok with removing the `-I` option in the current setting because structops already allows me to specify `--ignore` only once and then list the packages behind it.  I just read the clap documentation and it seems that you can also type `--ignore=package-a,package-b` right now. So I will be removing the `-I option` only.

vn971:
I tried to use the comma, it didn't work. (Actually I'd be surprised if it would work out of the box - after all, you might want commas in your arguments.)

Specifying multiple packages "with a space" (separate arguments) works, though. I'm still unsure if it's the best approach though. This way, if we'll want to have "subcommand arguments" in the future, like `rua upgrade a b c`, it will become harder. Because then you'll need to separate between `--ignore` arguments and the subcommand arguments, like `rua --ignore x y z -- a b c`. Actually it's not very realistic to have both of the the things in one invocation, but still.

If we, however, go with the comma approach, and only allow specifying `--ignore` once, then the command becomes clearer: `rua upgrade --ignore x,y,z a b c`. Does that make sense to you?

mqus:
Sure! Thats simple enough.

vn971:
Thanks for the contribution!

Issue #123, opened 2020-05-02
`--ignore`  is a required argument to `upgrade`


olantwin:
With the most recent release, `rua upgrade` fails:
```
error: The following required arguments were not provided:
    --ignore <ignored>
```

It seems like `--ignore` was accidentally added as a required argument rather than as an optional argument in #122 .

I'll have a look whether I can fix this, but I'm unfortunately not particularly familiar with rust and the way the arguments are parsed.

vn971:
Thanks for the bug report, I understood what's happening, will fix now.

vn971:
Fixed that, making a release now. Thanks again for the bug report!

olantwin:
Thanks for the quick fix!

Also learnt a thing or two while trying to fix it myself (and comparing my solution to your better one ;) )

vn971:
I think both solutions are good actually - just ended up committing mine because I already had a release in progress (and I like flat_map-s:). Cheers.

PR #124, opened 2020-05-02
Naive attempt at fixing #123

olantwin:
I made an attempt of fixing the issue I encountered.

As I am a beginner with rust, I would appreciate any feedback. Feel free to ignore my fix of course, if you think there's a better way to fix it.

Tested that this works with and without the `--ignore` flag (but since I didn't have any upgrades waiting, I couldn't test anything non-trivial).

### DIFF
--- a/src/cli_args.rs
+++ b/src/cli_args.rs
@@ -111,7 +111,7 @@ Supports: git, hg, bzr, svn, cvs, darcs. Currently by suffix only."
 			long = "ignore",
 			help = "Don't upgrade the specified package(s). Accepts multiple arguments separated by `,`."
 		)]
-		ignored: String,
+		ignored: Option<String>,
 	},
 }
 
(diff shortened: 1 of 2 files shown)


olantwin:
Fixed in dfd45415c5e3c63d0144d2b49e8f89c733a6fb5b .

vn971:
Hi olantwin, indeed I've already managed to make a fix as well, and a release on top of that, so I'll keep it.

Some comments if you might still be interested though:
* first, I think your solution is indeed correct, and works quite fine.:)
* I decided to go with a `flat_map` solution: approach the `Option` as a kind of collection in itself (containing 0 or 1 elements), then split each item of the collection by comma and derive a collection of already split packages to ignore. Essentially the same, just thought it might feel a bit more uniform.

That's about it.

olantwin:
Thanks for your explanation. Indeed, treating the `Option` as a collection seems very intuitive in this context.

PR #134, opened 2020-07-24
Update dependencies

Morganamilo:
The main point of this is to upgrade raur which now uses reqwest 10


vn971:
Nice, thanks! I also noticed it can even be improved a bit further, so did that as well, and merged now.

BTW, is the reqwest update something important? `cargo audit` doesn't warn me anything about it. Just curious. If it's something important, then maybe a release (for rua) is called for as well. WDYT ?

Morganamilo:
v10 adds async support. While raur doesn't support async because I'm waiting on async traits. This avoids having two separate versions of reqwest of its dependencies in my programs that use raur and async.

vn971:
Understood, thanks :+1:

PR #137, opened 2020-10-03
Fix clippy warnings

Morganamilo:


vn971:
Hey, thanks! Do you run clippy under a stricter settings than default to get these warnings? I have clippy as part of CI and manual release procedure, but I didn't see those warnings yet.

Morganamilo:
I use the defaults. But I am on nightly, so maybe they're new lints.

vn971:
Could be it. Thanks!

P.S. Also noticed that `cargo clippy -- -W clippy::pedantic` gives some more warnings, some of which I could agree to. Maybe the project could enforce those as well. (Will take a look at that after finishing my current WIP branch that I have locally)

Morganamilo:
I find pedantic to be a bit too far. Sometimes may run it just to see what it spits out and if i agree with any of the lints. But enforcing all of them I think is too far.

PR #138, opened 2020-10-07
Add support for BLAKE2 checksums

joanbm:
Their existence and usage are documented on ArchWiki:
https://wiki.archlinux.org/index.php?title=PKGBUILD&oldid=635970#b2sums

Adding support for them simply involves handling them similarly to the other already supported checksum types (sha1sums, sha256sums, etc.)

srcinfo.rs is updated since the old version didn't support BLAKE2 either

This has two effects:

* When building packages offline ('rua builddir --offline' or  'rua install --offline'), the BLAKE2 checksums are checked instead of ignored.
* No superfluous warnings are emitted on the b2sums variable on 'rua shellcheck'  on PKGBUILDs with BLAKE2 checksums, such as this one (Boost):  https://raw.githubusercontent.com/archlinux/svntogit-packages/41c5d32f4ff454ddfca19fcb4a8314594df453f6/trunk/PKGBUILD

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1295,9 +1295,9 @@ checksum = "6e63cff320ae2c57904679ba7cb63280a3dc4613885beafb148ee7bf9aa9042d"
 
 [[package]]
 name = "srcinfo"
-version = "0.3.2"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2dc5714b231447cde0c276ced9f6567c5068c6b48558d09cb004b077a27b4fe2"
+checksum = "a69c538e89460803500fca8d5f54a82f5ae1b3b2fc0661462ee41d8641a3571a"
 
 [[package]]
 name = "stacker"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ raur = { version = "3.0.1", default-features = false, features = ["rustls"] }
 regex = "1.3.9"
 rm_rf = "0.6.0"
 ruzstd = "0.2.1"
-srcinfo = "0.3.2"
+srcinfo = "1.0.0"
 structopt = "0.3.18"
 tar = "0.4.30"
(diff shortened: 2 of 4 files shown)


vn971:
Hi, that's nice, and I personally find b2 to be a good and valuable hash as well. (And regardless of personal preferences, it would make sense to support whatever's supported by Arch itself.)

I'll take a detailed look at the PR a bit later

vn971:
With regards to checking packages:

> When building packages offline ('rua builddir --offline' or 'rua install --offline'), the BLAKE2 checksums are checked instead of ignored.

Rua does not build packages completely on its own, and instead, still makes a `makepkg` invocation. If you know for a fact that b2sums are ignored by rua anyway, then it's strange and I'd like to investigate that. Do you observed this in your environment or was it a hypothesis? (I'll also check in my env.)

vn971 (approved):

vn971:
Thanks, merged! The question on skipping b2sums is still open to me

joanbm:
Hi! Thanks for merging. I was a bit unclear/misleading about what I meant by "When building packages offline, the BLAKE2 checksums are checked instead of ignored."

When building in offline mode, there are two makepkg invocations, one to download the sources and one to do the actual build. _The one to download the sources didn't check the BLAKE2 checksums_ - but did check other kinds of checksums. That's because rua creates a temporary PKGBUILD to download the sources and the BLAKE2 checksums weren't copied there unlike other kinds of checksums (in the code that's `download_srcinfo_sources` which calls `static_pkgbuild` where I made the change). However, when doing the actual build, the original PKGBUILD is used, which re-runs the checksum checks again, and there BLAKE2 checksums are actually checked, so they weren't skipped altogether. Example:

boost_rua_b2sums

So it's not terribly important that BLAKE2 checksums weren't checked while downloading the sources since they are re-checked during the build anyway, but it's nice to have them checked for consistency and so the build fails as early as possible.

Issue #145, opened 2020-11-26
Feature suggestion: clear cache


refaelsh:
Here is a scenario that happened to me. I was doing a `rua install foo-bar-package` (1 GB download) and the WiFi got disconnected in the middle of the download. This happened several times. I now have multiple fragments downloaded that are eating disk space. I suggest a `rua clean cache` command to get rid of them.

Thanks.

vn971:
@refaelsh hey! In the particular situation that you describe, duplication wouldn't (shouldn't) happen, because only the last build can ever leave garbage.

On the other hand, I imagine it could be nice to have the garbage cleared even for 1 build. I'd imagine two things could help:
* Have a command like `rua clean` (or `rua clean-cache`) to clean the build cache
* Automatically removing build garbage at the end of a successful build (with a flag to override it back if desired)

What do you think?

refaelsh:
Sounds great!

vn971:
I've added the tags to invite collaboration. (Personally busy with another task ATM, building the seccomp rules on non-x86_64 architectures.)

refaelsh:
Theoretically, if somebody would open a PR on this, what the command should do? Delete something, right? But what? This folder maybe: `~/.cache/rua/build/foo-bar-package/` or maybe `~/.cache/rua/build/`?

vn971:
To give an intermediate update, I'm super-busy with work ATM, but I'll really try to get back to it soon!

vn971:
Hey, sorry, that was a very slow "soon". I made it within 1 month though! :sweat_smile:  (was really busy with personal things and work)

I think `~/.cache/rua/build/foo-bar-package/` can be deleted at the end of a successful build and `~/.cache/rua/build/` can be deleted by `rua clean-cache`

That should be really straightforward, and an improvement upon the current situation. I think it's a really good thing to do

refaelsh:
Thank you for the reply. I will try and see wether my Rust skill are not rusted. If they are not, I will open a PR.

refaelsh:
One more question please. If `~/.cache/rua/build/` gets deleted, would it not break `rua upgrade`?

vn971:
@refaelsh shouldn't, `build` is only used during build, but not after. Git history of AUR itself is kept in `~/.config/rua/pkg/`, then copied inside `~/.cache/rua`, and the built artifacts are in `~/.local/share/rua`. This was done intentionally so that a user can delete the XDG cache and not lose anything important.

[...]

PR #146, opened 2021-01-06
Delete leftovers after a successful install

refaelsh:
This PR is in reference to #145.
Please be gentle :-)

P.S.
Made in Vim with Love.


src/action_install.rs
```diff
@@ -54,6 +56,10 @@ pub fn install(targets: &[String], rua_paths: &RuaPaths, is_offline: bool, asdep
 		is_offline,
 		asdeps,
 	);
+
+	let folder_deleter = Box::new(FolderDeleter::new());
+	let left_overs_deleter = LeftOversDeleter::new(folder_deleter);
+	left_overs_deleter.delete_folders(targets, rua_paths);
```
vn971, line 62: Hey, thanks for the PR!

One question on the libraries, what do you think of using `rm_rf` library for directory removal? It's a library that I created specifically to remove files and directories where `fs::remove_dir_all` would fail (it sometimes happens with AUR build dirs). This library is already used in a few places here, it's well tested as well. Then the only thing left would be to just call this function in a few places after the build process is completed and built artifacts are copied to `checked_tars` dir?

refaelsh: Thank you for being gentle! :-)
I will think over what you are suggesting, it looks logical. I will need a few days of meditation on that.

refaelsh: > Then the only thing left would be to just call this function in a few **places**

Why plural **places** and not singular place? As far as I understand from the code, its only a single place, right at the end of function `install` in file `action_install.rs`, correct?

vn971: @refaelsh Ah yes, indeed, you're right. Here it'll only be in one place, after the packages are moved to the `checked_tars` directory. It will also be needed for the potential `clean-cache` command, but that's a separate thing and can be added independently

refaelsh:
I tried using you `rm_rf` crate. I've change the `IFolderDeleter::delete_folder()` function return value to use your struct like this:
```
pub trait IFolderDeleter {
	fn delete_folder(&self, path: &PathBuf) -> rm_rf::Result<()>;
}
```
And I get the following error from the compiler:
```
   Compiling rua v0.17.18 (/home/refaelsh/repos/rua)
error[E0603]: type alias `Result` is private
  --> src/folder_deleter.rs:11:52
   |
11 |     fn delete_folder(&self, path: &PathBuf) -> rm_rf::Result<()>;
   |                                                       ^^^^^^ private type alias
   |
note: the type alias `Result` is defined here
  --> /home/refaelsh/.cargo/registry/src/github.com-1ecc6299db9ec823/rm_rf-0.6.0/src/lib.rs:4:5
   |
4  | use crate::error::Result;
   |     
```

Please advice.

vn971:
@refaelsh Sorry again for the slow reply, it seems that `rm_rf` library itself was missing proper "public" imports: https://github.com/vn971/rm_rf/commit/5d6694bf2b93dd629fbfac4399bdea13151b2ac8

If you change the rm_rf version to latest (0.6.1), the code should start compiling.

BTW, can you please explain a bit, what was the idea behind `IFolderDeleter`, can't we just get away with calling `rm_rf` directly? Or you wanted to gradually get away from it by first just replacing the implementation?

refaelsh:
> @refaelsh Sorry again for the slow reply,

This is perfectly ok :-) Thank you :-)

> If you change the rm_rf version to latest (0.6.1), the code should start compiling.

It compiles now.

> can't we just get away with calling `rm_rf` directly? 

We can. But we should not.

> BTW, can you please explain a bit, what was the idea behind `IFolderDeleter`

Its an interface. In Java/C# one would use the `interface` keyword and in C++ for example one would use an Abstract Class (since C++ lacks the `interface` construct). This approach has several advantages:
* It is Unit Testable (and I did indeed added unit tests for the `FolderDeleter` struct that implements the `IFolderDeleter` trait (interface).
* When in the future we will want to use something else instead of `rm_rf`, we will only have one place to change: the `FolderDeleter` struct.

vn971:
Ah, if we want to e.g. test it, then it makes sense indeed (e.g. "when you run install ABC with conditions XYZ, a call to remove directory KLM needs to be made"). Good stuff then! I'll read the new MR a bit later

refaelsh:
Question please. I wanted to see what/where is the CI/CD and I cant find it. The file `.travis.yml` cant be it because it does not contains `cargo build --release`. I also saw `release.sh`, but it also does not contains `cargo build --release`. Can you please point me in the right direction?

vn971:
Hey, took a while, it's always a bit harder to write reviews if they are not approving ones.

The main problem that I see with the PR as it's currently written, is what it aims to do and what it does.

If I understand the intent correctly, we want to introduce tests for package installation and directory deletion. I see that we're creating the traits for directory deletion, which many different structures (or mocks) could then implement. If I understand it correctly, the common approach, however, would be then to also use this abstract trait all the way up in the dependency tree with something like dependency injection.

For example, let's say we want to test how RUA works internally. Our "written in English" test would sound, I assume, smth like: "whenever a package is installed, a call to remove the build directory should be made". In terms of traits and mocks, it should be a "package installer" function that accepts a "directory deleter" trait, and calls it when required. Now we want to test the "package installer" function, and instead of giving the function a real "directory deleter", we would provide a fake one. We would then ask the "package installer" to proceed. The mock would tell us if "directory deleter" was indeed called in the process, and possibly check which arguments were used.

Going further into the dependency tree, the function that uses "package installer" could also take "package installer" as a trait, and in testing we could override that as well. That would be the full DI / testing approach.

Now we're getting to the problem. In this PR, "package installer" does not really accept a "directory deleter" trait. We're abstracting away from "directory deleter" by using a trait, but we can only ever use one specific "directory deleter", which is hardcoded in the current PR in line 60 of actions_install.rs: https://github.com/vn971/rua/pull/146/files#diff-863878877eeb9159c0270d8aa71de01d3281a4b761037e67af3430957626977fR60
We didn't test "package installer"-s behavior, we only tested a self-contained "directory deleter" behavior.

That's the problem that I saw, I'll write a separate response on where the code could go from there

vn971:
The solutions on how to address the problem, as far as I see, could be:

* Just use a direct call to `rm_rf` to remove the directory. We could then proceed with the actual feature of `rua clear-cache` and such. The imperative code for package installation would then remain untested as far as automated tests are concerned.

* Make "package installer" accept a "directory deleter". Said "package installer" should then be testable in actual automated tests. I'm not actually very convinced on what the point of the test would be. If the test would be to check that certain actions are called in certain order, then it would kinda literally repeat the code. The function at hand doesn't seem like a parser or any such function with a broad spectrum of possible arguments and broad spectrum of possible things that the function could do. I'm a big fan of testing, but only when we test something that is really different from literally repeating the code 1:1.

Note that if you'd like to proceed with the second approach, some other things would be good to fix:
* Think if we really need the `FolderDeleter` and `LeftOversDeleter` traits? It seems that a function `path: &PathBuf -> rm_rf::Result<()>` could replace the trait (or itself being a trait)? Or if "named traits" are preferred over "anonymous function", then maybe keep one of the traits since this is what we're actually interested in when we check if "package installer" works correctly?
* Note sure what `m_` denotes here, could be improved if we go for this approach: https://github.com/vn971/rua/pull/146/files#diff-3e1fe2772b5f94ba3578d10793d98ab1cd6ee248143221847a11dc9ded130e19R6
* Since the function returns a `Result` already, `unwrap()` at line 21 should be re-written to return actual values: https://github.com/vn971/rua/pull/146/files#diff-3e1fe2772b5f94ba3578d10793d98ab1cd6ee248143221847a11dc9ded130e19R21
* Panicking inside tests could be acceptable I guess, but it'd be still nice to provide written description for the reason of failure and why it's unacceptable for the test. This goes for the second `unwrap()` beind used and `assert_eq!`: https://github.com/vn971/rua/pull/146/files#diff-3e1fe2772b5f94ba3578d10793d98ab1cd6ee248143221847a11dc9ded130e19R54

vn971:
WRT testing, I've raised an issue here: https://github.com/vn971/rua/issues/151

vn971:
`release.sh` has `cargo publish`, which implies `cargo build --release`. This script, however, is for maintainers only, and normal user wouldn't benefit from that. The actual `cargo build --release` happens when you install `rua` as an AUR package here: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=rua#n33

Or when you follow README.md, choose the Rust type of installation and run `cargo install rua`: https://github.com/vn971/rua#install-the-rust-way

vn971:
Hope any of that helps...

refaelsh:
> Hey, took a while, it's always a bit harder to write reviews if they are not approving ones.

Thats ok, I completely understand you :-)
 
> The main problem that I see with the PR as it's currently written, is what it aims to do and what it does.
> 
> If I understand the intent correctly, we want to introduce tests for package installation and directory deletion. I see that we're creating the traits for directory deletion, which many different structures (or mocks) could then implement. If I understand it correctly, the common approach, however, would be then to also use this abstract trait all the way up in the dependency tree with something like dependency injection.

That would be te correct way - "all the way up in the dependency tree". I was just afraid to introduce such a big change, its a kind of a "revolution". But, if you don't have a problem with that, I am more then happy to go that way.

> For example, let's say we want to test how RUA works internally. Our "written in English" test would sound, I assume, smth like: "whenever a package is installed, a call to remove the build directory should be made". In terms of traits and mocks, it should be a "package installer" function that accepts a "directory deleter" trait, and calls it when required. Now we want to test the "package installer" function, and instead of giving the function a real "directory deleter", we would provide a fake one. We would then ask the "package installer" to proceed. The mock would tell us if "directory deleter" was indeed called in the process, and possibly check which arguments were used.

Understood and agreed.

> Going further into the dependency tree, the function that uses "package installer" could also take "package installer" as a trait, and in testing we could override that as well. That would be the full DI / testing approach.

That seems like a very big step for a single PR/branch. Maybe we should leave it for the future and/or a different PR that will be opened after that one is finished?

> Now we're getting to the problem. In this PR, "package installer" does not really accept a "directory deleter" trait. We're abstracting away from "directory deleter" by using a trait, but we can only ever use one specific "directory deleter", which is hardcoded in the current PR in line 60 of actions_install.rs: https://github.com/vn971/rua/pull/146/files#diff-863878877eeb9159c0270d8aa71de01d3281a4b761037e67af3430957626977fR60
> We didn't test "package installer"-s behavior, we only tested a self-contained "directory deleter" behavior.

Agreed. Again, I was afraid of making a "revolution".

> That's the problem that I saw, I'll write a separate response on where the code could go from there

refaelsh:
> Hope any of that helps...

It does! A lot! Thank you very much :-)

refaelsh:
> The solutions on how to address the problem, as far as I see, could be:
> 
>     * Just use a direct call to `rm_rf` to remove the directory. We could then proceed with the actual feature of `rua clear-cache` and such. The imperative code for package installation would then remain untested as far as automated tests are concerned.
> 
>     * Make "package installer" accept a "directory deleter". Said "package installer" should then be testable in actual automated tests. I'm not actually very convinced on what the point of the test would be. If the test would be to check that certain actions are called in certain order, then it would kinda literally repeat the code. The function at hand doesn't seem like a parser or any such function with a broad spectrum of possible arguments and broad spectrum of possible things that the function could do. I'm a big fan of testing, but only when we test something that is really different from literally repeating the code 1:1.
> 
> 
> Note that if you'd like to proceed with the second approach, some other things would be good to fix:

Yes, I would like to proceed with the second solution.

>     * Think if we really need the `FolderDeleter` and `LeftOversDeleter` traits? It seems that a function `path: &PathBuf -> rm_rf::Result<()>` could replace the trait (or itself being a trait)? Or if "named traits" are preferred over "anonymous function", then maybe keep one of the traits since this is what we're actually interested in when we check if "package installer" works correctly?

`LeftOversDeleter` is not a trait, its a concrete class that accepts a trait as a ctor parameter.

>     * Note sure what `m_` denotes here, could be improved if we go for this approach: https://github.com/vn971/rua/pull/146/files#diff-3e1fe2772b5f94ba3578d10793d98ab1cd6ee248143221847a11dc9ded130e19R6

'm_` is from the word member, as in data member of the class. Its a common (?) convention. I really like it. IMHO, it improves readability of the code. Are you not familiar with that convention?
 
>     * Since the function returns a `Result` already, `unwrap()` at line 21 should be re-written to return actual values: https://github.com/vn971/rua/pull/146/files#diff-3e1fe2772b5f94ba3578d10793d98ab1cd6ee248143221847a11dc9ded130e19R21

I don't understand.

>     * Panicking inside tests could be acceptable I guess, but it'd be still nice to provide written description for the reason of failure and why it's unacceptable for the test. This goes for the second `unwrap()` beind used and `assert_eq!`: https://github.com/vn971/rua/pull/146/files#diff-3e1fe2772b5f94ba3578d10793d98ab1cd6ee248143221847a11dc9ded130e19R54

I will add it if you want to. But, from my experience, when I get an email from CI/CD server, eventually I anyway get my hands dirty to investigate why the unit test failed, so the failure message does not really matters.

refaelsh:
> If the test would be to check that certain actions are called in certain order, then it would kinda literally repeat the code.

Yea, good point. I will take it under advisement.

refaelsh:
@vn971 I am done with what you suggested and I feel better :-) The "revolution" is not that big as I feared.

vn971:
To give an overall comment while I see you're making commits in `git`. The most valuable places to cover with tests should probably be those where the logic flow itself is non-trivial. E.g. if the code is a step-by-step direct execution, then there's not much to test there except to repeat the same sequence. If the code loops back and does weird things like that, it's a good candidate to mock what it touches and test that the order is good.

For example, recursive package download `recursive_info` could be tested to make few requests with no extra unneeded packages. (Although here I'd feel a little bit guilty at not accepting Morganamilo-s PR back in the days, but the approaches were really different there.)

Otherwise I don't actually see places that wouldn't directly say what happens and how TBH. It's all really "declarative" code visually and maintenance-wise, even though technically "imperative" all the way through.

vn971:
If the aim is to improve code quality in general, then CI would be a very attacking point to my eyes as well. One good approach could also be to reduce the usage of `panic`/`unwrap` (I'm a bit guilty of doing that 2 years ago) and use `Result`. This could unblock us to e.g. do https://github.com/vn971/rua/issues/148

refaelsh:
> while I see you're making commits in `git`

No no, that is it, I am done. Ready for you reviewing pleasure :-)

vn971:
Hey! I think I'm going to close the PR, it overall seems to increase the conceptual complexity (e.g., number of concepts) of the project without giving a lot of the benefits that this could introduce. I'll leave more comments to the exact files around

.gitignore
```diff
@@ -1 +1,5 @@
+# For Vim users
+.nvimrc
+.vimrc
```
vn971, line 3: Note: this is fine to have both in project-s .gitignore, and in your personal in `~/.gitignore` or `~/.config/git/ignore`.

Some people even [advocate to always keep your personal editor's settings in your personal files](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/). I think this is a cleaner practice indeed

src/left_overs_deleter.rs
```diff
@@ -0,0 +1,82 @@
...
+
+		let targets = ["some_target".to_string()];
+		let rua_paths = RuaPaths::initialize_paths();
+
+		let res = left_overs_deleter.delete_folders(&targets, &rua_paths);
+		match res {
+			Ok(()) => (),
+			Err(_error) => panic!(),
+		}
+	}
```
vn971, line 81: I've decided to go with this simple PR: https://github.com/vn971/rua/pull/185/files

The reason for doing that is that the current code with `LeftOversDeleter`, while mock**able**, does not actually strongly use its mocking ability. E.g. what's done, is that we ask the `LeftOversDeleter` to delete something and we check that this was done.

What we could want, is to invoke `install` and check that deleter was invoked with the right arguments.

Additionally, this particular project is very much declarative. Meaning that the code literally says in which order to execute which actions. Tests are not as good for testing such behavior as they would literally repeat the very same things. I'd subjectively say that tests are the best when they test code where it's really easy to make a mistake or get the logic straight, e.g. code involving loops, recursion, state-dependent logic, timeouts etc. But not necissarily code that says "delete" and does "delete". That's my subjective opinion as the maintainer, anyway...

rsheink:
Thank you very much for your time @vn971 :-)
I had much fun reading your remarks. Learned a lot from them.

refaelsh:
Thank you very much for your time :-)
I've had lots of joy reading your remakes. I've learned a lot from them.

refaelsh:
> Meaning that the code literally says in which order to execute which actions. Tests are not as good for testing such behavior as they would literally repeat the very same things.

I am now at a phase of my professional career where I understand what you are saying here. It was not the case when I opened the PR :-)

refaelsh:
>> That's my subjective opinion

I agree here, code that has no `ifs` and/or `fors` and/or `whiles` should not be unit tested due to what you said: `literally repeat the very same things`. This is similar to what [Martin Fowler](https://martinfowler.com/) calls the `Humble Object` pattern, and he clearly states this pattern does not requires Unit Tests (most of the time anyway).

PR #159, opened 2021-05-03
Add toolchain file

Morganamilo:
This sets the default toolchain to stable without having to set an env
var in the pkgbuild.


vn971:
Thank you! :heart:

PR #160, opened 2021-05-26
FEATURE: Document ccache usage

oleid:
I just figured out how to use `ccache` with `rua`.  As I didn't find an issue asking about it or anyone having ever tried it, I thought it might be worthwhile to create some documentation on the subject.

What would be the correct place to link the documentation so that it can be found?  EDIT: Linked it to main `README.md`


vn971:
@oleid I've made two micro-changes and now merged it. Thanks for documenting it!

PR #161, opened 2021-06-01
Add colors to review menu

intgr:
With colors it's much easier to tell, where one menu option ends and the next one begins.

The color scheme is based on a simple "traffic lights" concept:
* green is safe (shellcheck/diff)
* red may be dangerous (use package)
* yellow is in between (merge)

Additionally:
* "run terminal" is cyan, to distinguish it from other options
* inactive options appear gray.

In my testing it works well with both white and black terminal background.

### DIFF
--- a/src/reviewing.rs
+++ b/src/reviewing.rs
@@ -4,6 +4,7 @@ use crate::terminal_util;
 use crate::wrapped;
 use log::debug;
 use std::path::Path;
+use colored::Colorize;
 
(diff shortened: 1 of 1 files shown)


intgr:
Screenshots of the result:

image

image

vn971:
Interesting, thanks! I wonder how it plays out with the SUID warning, let me briefly test that out before merging

vn971:
OK, this PR doesn't actually change "tar_review" stage, only package review at build stage. I think it's good, thanks again for the detailed commit message & the screenshots, plus a nice small PR! :raised_hands:

PR #162, opened 2021-06-01
Split out libalpm as a cargo feature

vn971:
Split out libalpm as a cargo feature

Splitting out libalpm as a feature would allow
removing libalpm as a link-time dependency.

This will in turn make `rua` work on both "git" version of pacman,
and not break when pacman upgrades itself (via libalpm).


vn971:
Opinions @Morganamilo @VukoDrakkeinen  or anyone else?

Morganamilo:
I don't see the use in a feature, just pick one aproch and just use that.

vn971:
Merged for now, will keep both versions for now to have some flexibility in choosing [what to use]. In the future might bias towards one of the solutions, but for now it's good

Morganamilo:
alpm.rs now has a static feature. Not that I'd exactly recommend enabling it for general use. It's there if you want to play around with it.

Issue #126, opened 2020-05-17
seccomp-armv7l.bpf: No such file or directory


diegoflorez:
I'm trying to use rua on raspberry-pi 4, but I keep getting this error when I try to install any aur package using `rua install package-name` command.
```/home/user/.config/rua/.system/wrap.sh: line 11: /home/user/.config/rua/.system/seccomp-armv7l.bpf: No such file or directory```
Although if I go to `~/.config/rua/pkg/package-name/` and run `makepkg -si` the package is correctly installed as indicated in `[AUR wiki page](https://wiki.archlinux.org/index.php/Arch_User_Repository)`

Does anyone have any idea why this may be happening and how to fix it?

vn971:
@diegoflorez hi, thanks for raising the issue! So this day has come, rua was tried to be run on an `arm` device.:) The problem observed is because currently, seccomp bpf filters were generated once for i686 and x86_64 architectures. I've mentioned that some time ago here: https://github.com/vn971/rua/issues/91#issuecomment-559064460

Now, however, it would be nice to auto-generate the seccomp pbf filters for other platforms automatically.. I'll take a look at it somewhat later (probably next week).

diegoflorez:
Great job with rua, by the way. And Thanks! :wink:

zegelin:
Running into this issue too while trying to use RUA on a Raspberry Pi 4.

I ran `seccomp-gen` and copied the resulting file to `.config/rua/.system/seccomp-armv7l.bpf`, but it appears that the entire `.system` directory gets nuked on every run.

Is there any way to use this BPF file?

vn971:
@zegelin yeah, I believe you also wrote a comment on AUR yesterday evening, right?

Anyway, I agree with everything that you wrote basically. Do you have time/energy to write a PR? I see two choices:

1) create the seccomp file in AUR, for not out-of-the-box architectures. And don't nuke (as you said:) the system directory. Theoretically it leaves some possibility to get few "garbage" files there, but I guess that's all right in a way. (This approach will also require touching up README.md)

2) the other approach is to integrate libseccomp into rua itself, link against it same as that programin `C` does.

I kinda like the 1-st approach because the file is really static and never changes I think, but I'm open to opinions.

zegelin:
> @zegelin yeah, I believe you also wrote a comment on AUR yesterday evening, right?

Nah, not me. I discovered this project today.

[...]

PR #167, opened 2021-06-06
Generate seccomp.bpf during build

bbx0:
This is a proposal for a runtime independent handling of the `seccomp.bpf` file. This allows support for additional architectures like `aarch64` or `arm` (see #126).

#### How does it work?
* A specific `seccomp.bpf` is generated via `build.rs` during compilation.
  * The generated `seccomp.bpf` matches the `target_arch` of the `rua` binary. It's no longer required to distinguish different versions during runtime.
  * The `seccomp.bpf` is embedded to the binary, just like the prebuilt ones before. During runtime it is extracted to `.system/seccomp.bpf`. (The prebuilt files are removed by this pull request.)
* A new *build-time* dependency `libscmp` is introduced to generate the `seccomp.bpf` file.
  * There are various crates available for that purpose but to me `libscmp` seems most appealing. It has a very low footprint and provides all required features (especially `seccomp_syscall_resolve_name` and `seccomp_export_bpf`). An alternative would be the `libseccomp` crate.
 
#### How can it be tested / reviewed?
* The implementation exports next to the `seccomp.bpf` a `seccomp.pfc` to `OUT_DIR` containing the human-readable output of the filter. (This second file is not embedded to the binary but can easily be inspected.)
* The `md5sum` for the `seccomp.bpf` file generated by this implementation is identical to the old one generated by [seccomp](https://github.com/kurnevsky/dotfiles/blob/master/.seccomp/) (https://github.com/kurnevsky/dotfiles/commit/1432bf8ae2b62f16142b7c87ad2e0fd28d27e9fc).
```sh
$ find . -name \*.bpf -exec md5sum {} \;
de85db9cf299f131c51cecee5f84ce77  ./res/seccomp-i686.bpf
f9cd17bc2fe18abc4f672071dd9e0c9c  ./res/seccomp-x86_64.bpf
de85db9cf299f131c51cecee5f84ce77  ./target/i686-unknown-linux-gnu/debug/build/rua-5b88deb180f14f08/out/seccomp.bpf
f9cd17bc2fe18abc4f672071dd9e0c9c  ./target/x86_64-unknown-linux-gnu/debug/build/rua-bdd5a199bf2717e3/out/seccomp.bpf
```

#### Remarks
* The `seccomp.bpf` file can contain the ruleset for all supported architecures at once. This comes with an increased file size but may reduce the logic in `build.rs` by a few lines. (I don't think it's worth it.)

I hope this helps. In case another solution fits better, I'm happy to revise.


vn971:
Hey @bbx0, this is actually a super-useful MR, sorry for only starting on that now. I'm looking at the code now, thinking on how to integrate it. Some things that I need to check are:

* Do we need to update README.md to ask for `libseccomp` installation? (Likely yes.)
* The AUR packages need to be updated to include libseccomp as well
* `build.rs` needs to be code-formatted

No action required for now, just wanted to let know that I've started on this..

vn971:
@bbx0 I have a question, can you please give a little bit of insight on why you included `name_to_handle_at` into the list of forbidden syscalls? From what I'm reading (`man 2 name_to_handle_at`), this seems fairly innocent, e.g. something that a legit build could do. Does this syscall have a history of being easy to exploit, or some other things about it?

build.rs
```diff
@@ -1,17 +1,127 @@
...
+			"kexec_load",
+			"keyctl",
+			"lookup_dcookie",
+			"mbind",
+			"nfsservctl",
+			"migrate_pages",
+			"modify_ldt",
+			"mount",
+			"move_pages",
+			"name_to_handle_at",
```
vn971, line 80: @bbx0 I have a question, can you please give a little bit of insight on why you included `name_to_handle_at` into the list of forbidden syscalls? From what I'm reading (`man 2 name_to_handle_at`), this seems fairly innocent, e.g. something that a legit build could do. Does this syscall have a history of being easy to exploit, or some other things about it?

vn971: UPD: I see now that this is also now used in the upstream code that you've referenced. Same with `chroot`. I'll check there

vn971:
@bbx0 Strange, I thought I commented here? Anyway, thanks a lot for the PR! I've made few changes to e.g. disable a compilation warning and some style fixes, it's on `master` now and released as well.
Thanks!

build.rs
bbx0: @vn971 Yes, I tried to reproduce the existing behavior of rua, which did not use the most recent version of upstream (but referenced commit). The current version of upstream seems also not deny `name_to_handle_at` or `chroot` anymore.

bbx0:
Thanks a lot for pulling this in! Glad it was helpful. Running already the new release on my rp4 (aarch64). :)

Issue #170, opened 2021-06-24
Pkgbuild still uses the git feature


Morganamilo:
the pkgbuild still uses the git feature when pacman-git is installed causing it to fail to build because the feature is now libalpmgit.

Though it should just be removed altogether as the default is to shell out to pacman and that works for all pacman versions.

vn971:
@Morganamilo Thanks for noticing/writing! Fixed it now (pushed a new commit). I've left the pkgrel unchanged because I don't want people to re-build the tool if they have a working version already anyway. Not 100% if that's the right call tho, but leaving it as-is for now

PR #171, opened 2021-07-02
Remove git feature mention from readme

micwoj92:
See https://github.com/vn971/rua/issues/170


vn971:
@micwoj92 Thanks for spotting and for the MR! Makes total sense

Issue #106, opened 2020-03-02
Emulate or allow passing `--assume-installed` to pacman


OliverUv:
I've installed rustup, cargo, etc through "the rust way" with rustup.

Now I want to install wasm-pack with rua. I can download the MAKEPKG of AUR's wasm-pack and remove cargo from the dependency list, but I would prefer to handle all my aur installs through rua.

I'd like to be able to do something like `rua install --assume-installed rust=1:1.41.0-1 wasm-pack`, or be able to add rust/cargo to some list of assumed installed packages in a conf file.

I may be mistaken in how exactly to use the `--assume-installed` argument for pacman as I haven't needed to use it yet, so the command above is only an example.

Adding `IgnorePkg = cargo` or `IgnorePkg = rust` to pacman.conf does not seem to have any effect.

vn971:
Hi Oliver, thanks for raising! Would you be willing to work on that feature?

> Adding IgnorePkg = cargo or IgnorePkg = rust to pacman.conf does not seem to have any effect.

This section does not silently omit a dependency in pacman either I think, so you'll have to explicitly ignore the dep if you want to install a package.

Also note that rua is unaware of versions at the moment, so it'll ignore the whole package, not just a specific version of it. https://github.com/vn971/rua#limitations

That being said, I think it's certainly doable. During package resolution, you should just filter out dependencies provided by `--assume-installed`.

What do you think?

OliverUv:
I'm currently busy but will check out the rua code base next week and get
back to you.

PR #174, opened 2021-08-11
Add assume-installed #106

The-Corrupted:
Added the ability to specify packages that should be filtered out from the pacman_deps list. makepkg doesn't have a --assume-installed equivalent so if a filter is provided by the user then nodeps is added to the makepkg arguments list.

First time contributing to a rust project. Any feedback is welcome.

### DIFF
--- a/src/action_builddir.rs
+++ b/src/action_builddir.rs
@@ -6,7 +6,7 @@ use std::path::Path;
 use std::path::PathBuf;
 
 /// Build and install a package, see `crate::cli_args::Action::Builddir` for details
-pub fn action_builddir(dir: &Option<PathBuf>, rua_paths: &RuaPaths, offline: bool, force: bool) {
+pub fn action_builddir(dir: &Option<PathBuf>, rua_paths: &RuaPaths, offline: bool, force: bool, no_deps: bool) {
 	// Set `.` as default dir in case no build directory is provided.
 	let dir = match dir {
 		Some(path) => path,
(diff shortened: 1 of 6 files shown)


The-Corrupted:
> What'st the point of this if nodeps is going to be passed to makepkg? Wouldn't it make sense to just implement it as nodeps?

Well there appears to be two sets of dependency checks happening. One happens within rua and the other happens from makepkg. This patch filters out dependencies on ruas side so that the user isn't prompted to download a package they may already have ( tested by installing paru so in that case, cargo ). Nodeps is then passed to makepkg under the assumption that there aren't additional dependencies in makepkg that rua didn't catch. 

I agree that passing nodeps isn't great, however to get this working without passing nodeps you'd have to modify depends and makedepends on the users behalf or have the user go in and manually remove packages from PKGBUILD which is what this was meant to address. I choose the easier option which is to that all dependencies were found ( or ignored ) on rua side and to skip makepkgs dependency checks. All of that said, I'm open to investigating other solutions.

Morganamilo:
Ah right didn't realize rua didnt resolve the repo deps itself instead of using -i

vn971:
@The-Corrupted any reasons for closing it? I actually wanted to get back on this in the coming days. Is this no longer relevant, or you lost interest in maintaining the PR / something else?

The-Corrupted:
@vn971 No, sorry. I was cleaning up my github. It was late and I wasn't paying attention. I've reopened it.

vn971:
Thanks for the reply! I was so far this week meditating on the code a little bit, not exactly immediately grasping the need, or rather, whether this specific approach would be what users want to have the most.
The need itself is completely valid of course, I'm pretty sure it exists, just have to dive into it again and understand the best approach to handle this

PR #186, opened 2022-04-03
WIP: Fix 'builddir install bug' (#165) and reuse existing pacman logic

vn971:


vn971:
@raphCode Thoughts? This MR incorporates both your changes and addresses my latest remark [here](https://github.com/vn971/rua/pull/173#issuecomment-1086832895)

vn971:
UPDATE: on the other hand, the action `or install manually and enter M when done` gets broken this way. I'll have to think a bit more on what behavior I even want from the tool, exactly...

PR #193, opened 2022-07-08
Adds the `native-tls` feature

xaocon:
Allows a user to build with TLS supported by the native platform.
Especially useful when you need to use your own CA cert.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -230,12 +230,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "core-foundation-sys"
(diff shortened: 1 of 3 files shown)


vn971:
Hi, thanks for the MR! Was super-loaded with work for a while, only found some time to review now sorry.

Unrelated, with support for native TLS now merged, I'll still default to rustls for now. While not as flexible and a debatable decision regardless, rustls is also not vulnerable to e.g. the openssl CVE from November this year. Just sharing some parallel thoughts here

Thanks again!

Issue #65, opened 2019-09-10
asdeps and explicit are not handled properly


Morganamilo:
Currently rua uses a pretty naive solution to figure out is a package should be installed explicitly.

There's currently two problems with it:
 
1. It assumes an entire pkgbase is either asdeps or explicit. This it not the case, for example if I install `libc++` then `libc++abi` will get pulled in as a dependency.  Currently rua will suggest them both be explicit.

2. Install reason should not change when upgrading a package. Say I already have `foo=1` installed explicitly, then I want to install `bar` and `bar` depends on `foo=2`. Then rua would pull that in as a dependency and suggest I install it with `--asdeps`, changing the install reason.

Do note due to rua not handling versioned deps I think point 2 is unreachable because rua would consider the dep `foo=2` satisfied even though it should not be.

vn971:
Since both potential problems are apparently not touching the tool, closing.

Morganamilo:
Both problems? 1 is still valid

```
morganamilo@Octavia ~ % MAKEPKG_DUMMY_BUILD=1 rua install libc++
Package libc++ depends on libc++abi. Resolving...

In order to install all targets, the following pacman packages will need to be installed:

And the following AUR packages will need to be built and installed:
  libc++abi
  libc++

Proceed? [O]=ok, Ctrl-C=abort. o
WARNING: your AUR repo is up-to-date.
If you continue, the build directory will be removed and the build will be re-run.
If you don't want that, consider resolving the situation manually,
for example:    rua builddir "/home/morganamilo/.cache/rua/build/libc++"

Reviewing "/home/morganamilo/.config/rua/pkg/libc++". 
[S]=run shellcheck on PKGBUILD, [D]=(identical to upstream, empty diff), [T]=run shell to edit/inspect, [O]=ok, use package t
Exit the shell with `logout` or Ctrl-D...
morganamilo@Octavia master ...rua/pkg/libc++ % vim PKGBUILD 
morganamilo@Octavia master ...rua/pkg/libc++ % 
Reviewing "/home/morganamilo/.config/rua/pkg/libc++". 
[S]=run shellcheck on PKGBUILD, [D]=view your changes, [T]=run shell to edit/inspect, [O]=ok, use package o
==> Making package: libc++ 8.0.1-1 (Mon 16 Sep 2019 21:55:04 BST)
==> Checking runtime dependencies...
==> Checking buildtime dependencies...
==> Entering fakeroot environment...
==> Tidying install...
[...]

PR #208, opened 2023-05-31
Do not change install reason during upgrades

### LINKED ISSUE #169: Rua upgrading packages as dependency
First of all, I'm very new to open source contributing, so it's possible that I make a stupid mistake somewhere. Please bear with me.

Basically, the problem I encounter is that when upgrading or reinstalling, rua seems to mark packages as dependency when it shouldn't. I found this out because Pamac lists several of my AUR packages as orphans.

For example, [Zotero](https://aur.archlinux.org/packages/zotero). Pamac lists it as an orphan, and when I `pacman -Qi zotero | grep Reason`, I indeed get `Install Reason  : Installed as a dependency for another package`. But I never marked it as such myself.

When I update or reinstall a package through rua, it does a simple `pacman -U [package.tar.zst]`, so it doesn't explicitly tell Pacman to make it a dependency (as was the case in issue #2). And yet it somehow ends up as a dependency. When I repeat the same `pacman -U` command manually and add `--asexplicit`, the package is marked as explicitly installed again. My naive solution therefore would be to add `-asexplicit` to the Pacman commands when reinstalling or upgrading.

When I install a new package with `rua install [package]` it is marked as explicitly installed (as it should be), but as soon as I update or reinstall, the package becomes marked as a dependency.

I don't think Pamac can be interfering here, because its AUR option is disabled and I only use it to view installed packages.

InvalidUsernameException:
When upgrading AUR packages `rua` always adds the `--asdeps` switch to the `pacman` command line. That means that packages that were previously installed explicitly most likely end up as orphans because they have no other packages depending on them. This is annoying because it means one needs to run `pacman -D --asexlict <list of upgraded packages>` after every upgrade of AUR packages to keep the install reasons correct and to prevent accidental uninstallation.

This PR changes the behavior during upgrades to not include the `--asdeps` switch which means the install reasons of the upgraded packages stay unchanged.

This should fix #169, but not the related #65, which seems different despite someone having commented that the issues are duplicates of each other.

### DIFF
--- a/src/action_upgrade.rs
+++ b/src/action_upgrade.rs
@@ -70,7 +70,10 @@ pub fn upgrade_real(devel: bool, rua_paths: &RuaPaths, ignored: &HashSet<&str>) 
 			let user_input = terminal_util::read_line_lowercase();
 			if &user_input == "o" {
 				let outdated: Vec<String> = outdated.iter().map(|o| o.0.to_string()).collect();
-				action_install::install(&outdated, rua_paths, false, true);
+				// Using `false` here means we pass neither `--asdeps` nor `--asexplicit`
+				// to pacman, effectively leaving install reasons untouched.
+				let asdeps = false;
+				action_install::install(&outdated, rua_paths, false, asdeps);
 				break;
 			} else if &user_input == "x" {
 				break;


InvalidUsernameException:
@vn971 Since you mentioned somewhere else that your email notifications weren't working in the past: Any chance you can take a look at this PR if you haven't seen it yet?

PR #225, opened 2024-08-02
Use --needed flag for pacman installation in README

emilyinspace:


emilyinspace:
I thought it might be a good idea, since the packages might already be installed

vn971:
Sorry for the late merge :see_no_evil: Thanks for the PR!

